PR #24, opened 2026-05-09
feat(docker): single-command Docker stack for local development

andrefrd:
## Summary

- Adds a self-contained `docker compose up --build` workflow that brings the full Laravel + Vite + Reverb + Horizon + scheduler stack online inside one container, plus Postgres 16, Redis 7, and Mailpit. No host PHP / Node prerequisites.
- Replaces the existing `compose.yaml` (a stale Sail file pointing at `./vendor/laravel/sail/runtimes/8.5/` — Sail isn't installed and the path doesn't exist) with a self-describing build that uses a new `docker/Dockerfile`.
- The same `docker/Dockerfile` exposes a `production` target so self-hosters get a hardened image (composer `--no-dev`, npm production build, SSR build, OpCache locked) from the same source contributors use in dev.
- Documents the workflow under "Get Started" in the README — common commands, override mechanism, troubleshooting, and the production build hint.

## What changed

| File | Purpose |
|---|---|
| `docker/Dockerfile` | Multi-stage: shared `system-base` (PHP 8.4-FPM Alpine + Postgres / Redis / intl / sockets / redis extensions) → `composer-deps` / `composer-deps-prod` → `wayfinder-gen` → `asset-build` → `dev` / `production` targets. UID/GID build args. |
| `docker/{nginx,php.dev,php.prod,supervisord.dev,supervisord.prod}.{conf,ini}` | Sidecar configs. Dev opcache validates timestamps for hot-reload; prod is hardened. Supervisord runs fpm + nginx + reverb + horizon + scheduler (+ vite in dev). |
| `docker/entrypoint.sh` | Idempotent: seeds `.env` from `.env.docker.example`, generates `APP_KEY`, waits for Postgres, runs migrations, links storage, generates Passport keys, regenerates Wayfinder helpers, and re-installs deps if anonymous volumes were wiped. |
| `docker/.env.docker.example` | Docker-tuned env template (service-name hosts, sane defaults). Does not touch `.env.example`, so the existing Herd workflow is untouched. |
| `docker/postgres-init.sh` | Creates the `trypost_test` database `phpunit.xml` expects, on first Postgres boot. |
| `compose.yaml` | Self-contained dev stack. Bind-mounts source for hot reload; named volumes for `vendor/` and `node_modules/` so image-baked deps survive. Healthchecks gate `app` startup until Postgres / Redis are ready. Sets `DB_HOST=pgsql`, `DB_USERNAME=postgres`, `REDIS_HOST=redis` as process env so `.env.testing`'s pre-existing CI defaults don't break in-container tests. |
| `compose.override.yaml.example` | Per-developer customizations (UID/GID alignment, port conflicts, Xdebug, Selenium for Dusk). |
| `.dockerignore` | Keeps the build context small and prevents secrets / dev artifacts from leaking into the image. |
| `.gitignore` | Adds `.env.testing.local` as a safety net for any per-developer testing override. |
| `README.md` | New "Run with Docker" section covering single-command boot, common commands, override workflow, troubleshooting (port conflicts, UID on Linux, first-boot expectations), and the production target hint. |

## Why these specific choices

- **Single container with supervisord** instead of separate compose services for fpm / nginx / reverb / horizon / scheduler / vite, because the `@laravel/vite-plugin-wayfinder` plugin shells out to `php artisan` at runtime — splitting Vite into a Node-only container would break it. Co-locating mirrors the production supervisord layout and gives one mental model.
- **Postgres 16-alpine** matches `.env.ci`. `docker/postgres-init.sh` covers the `trypost_test` database that `phpunit.xml` pins.
- **`.env.docker.example` separate** from `.env.example` so Herd / `composer dev` users keep their flow unchanged. The entrypoint only seeds the Docker variant when `TRYPOST_DOCKER_BOOTSTRAP=1` (set by compose).
- **UID/GID build args** (default 1000) keep bind-mount writes owned by the host user. Linux contributors on non-1000 systems pass `UID=$(id -u) GID=$(id -g)`.

## Test plan

- [x] `docker compose up --build` from a fresh clone reaches healthy in ~3-4 min on first boot, seconds on subsequent boots.
- [x] `curl -fsS http://localhost:8000/up` → 200.
- [x] Homepage at `http://localhost:8000` renders (302 → /login).
- [x] Vite dev server reachable at `http://localhost:5173/@vite/client` → 200.
- [x] Mailpit UI at `http://localhost:8025` → 200.
- [x] `docker compose exec app php artisan test --compact tests/Feature/Auth/PasswordResetTest.php` → 5 passed (10 assertions).
- [x] `docker compose exec app php artisan horizon:status` → running.
- [x] `docker compose exec app php artisan schedule:list` → 4 scheduled commands.
- [x] `docker compose down && docker compose up -d` → fast restart, no re-migration thanks to `migrate --graceful`.
- [x] `docker compose down -v && docker compose up --build` → full reset works; `composer install` + `npm ci` re-run inside the container.
- [x] `docker build --target production -t trypost:prod -f docker/Dockerfile .` → production image builds cleanly without source-mount.
- [x] `compose.yaml` validated: `docker compose config --quiet` exits 0; `docker buildx --check` reports no warnings.

## Out of scope

- Production deploy guides (Swarm / Kubernetes / Forge). The `production` target is offered, but orchestration is left to self-hosters.
- Selenium / Dusk service: omitted from `compose.yaml` to keep first-boot small. `compose.override.yaml.example` shows how to add it.
- CI changes: `.env.ci` and `.github/workflows/` are unchanged.


docker/.env.docker.example
```diff
@@ -0,0 +1,177 @@
...
+AI_TEXT_MODEL=gpt-5.4
+AI_IMAGE_PROVIDER=openai
+AI_AUDIO_PROVIDER=elevenlabs
+
+# Stripe (only required when SELF_HOSTED=false)
+STRIPE_KEY=
+STRIPE_SECRET=
+STRIPE_WEBHOOK_SECRET=
+
+STRIPE_STARTER_MONTHLY=
```
paulocastellano: @andrefrd for self hosted plans not are required, can you remove that?

docker/.env.docker.example
```diff
@@ -0,0 +1,177 @@
...
+ANTHROPIC_API_KEY=
+GEMINI_API_KEY=
+ELEVENLABS_API_KEY=
+
+AI_TEXT_PROVIDER=openai
+AI_TEXT_MODEL=gpt-5.4
+AI_IMAGE_PROVIDER=openai
+AI_AUDIO_PROVIDER=elevenlabs
+
+# Stripe (only required when SELF_HOSTED=false)
```
paulocastellano: @andrefrd same here.

docker/php.dev.ini
```diff
@@ -0,0 +1,22 @@
+; TryPost — PHP settings for local development.
+memory_limit = 512M
+upload_max_filesize = 100M
```
paulocastellano: @andrefrd we need support up to 1GB, for videos.

docker/php.prod.ini
```diff
@@ -0,0 +1,20 @@
+; TryPost — PHP settings for production.
+memory_limit = 256M
+upload_max_filesize = 100M
```
paulocastellano: @andrefrd same here, we need support 1GB

README.md
```diff
@@ -99,6 +99,69 @@ Tired of paying $30–$200/month per user for a scheduler? Tired of your content
 | **Self-host** | Free forever. [Installation guide &rarr;](https://trypost.it/docs/getting-started/installation) |
 | **Run with AI assistants** | [Connect Claude / Cursor / ChatGPT via MCP &rarr;](https://trypost.it/docs/mcp) |
 
+## Run with Docker (recommended for contributors)
```
paulocastellano: @andrefrd we need remove it from README, because we have a dedicated repository for docs:

https://github.com/trypostit/trypost-docs

paulocastellano (changes requested):

docker/.env.docker.example
andrefrd: Done, dropped the whole STRIPE_* block in d12f8d3 since the Docker stack defaults to SELF_HOSTED=true.

docker/.env.docker.example
andrefrd: in same commit, d12f8d3 removes the price IDs alongside the Stripe credentials.

docker/php.dev.ini
andrefrd: Done in d12f8d3, upload_max_filesize, post_max_size, and nginx client_max_body_size all bumped to 1G.

docker/php.prod.ini
andrefrd: Same commit, d12f8d3 mirrors the change in php.prod.ini.

README.md
andrefrd: Done in 5fb8b85, this section wsas removed. 
I can open a PR against trypost-docs with the rewritten guide if useful, lmk

andrefrd: was*

paulocastellano: @andrefrd please do it, send a PR to trypost-docs.

PR #47, opened 2026-05-19
Fix: JSON columns with default('[]') fail on MySQL 8

sebastian-works:
## Problem

When running TryPost on MySQL 8, the migrations fail with:

```
SQLSTATE[42000]: Syntax error or access violation: 1101 BLOB, TEXT, GEOMETRY or JSON column 'media' can't have a default value
```

MySQL does not allow string literal defaults on JSON columns, regardless of version. This affects anyone self-hosting TryPost with MySQL instead of PostgreSQL.

## Changes

Replaced `->default('[]')` with `->nullable()->default(null)` on two JSON columns:

- `media` column in `create_posts_table`
- `reactions` column in `create_post_comments_table`

## Testing

Verified that both migrations run successfully on MySQL 8.0.45 (Ubuntu 24.04, aarch64).

## Notes

This does not affect PostgreSQL installs — `nullable()->default(null)` works correctly on both databases. Application code treating `null` as an empty array is standard Laravel behavior.

Related to the MySQL support advertised in the docs — without this fix, MySQL installs cannot complete the migration step.


app/Http/Controllers/App/AssetController.php
```diff
@@ -145,7 +145,7 @@ public function storeFromUrl(StoreAssetFromUrlRequest $request, UnsplashService
         $filename = Str::uuid().'.'.$extension;
         $path = 'medias/'.$filename;
 
-        Storage::put($path, $response->body());
+        Storage::disk('public')->put($path, $response->body());
```
paulocastellano: @albertvisuals we can't setup a fixed disk, because you are free to use any disk like s3, r2 and etc.

app/Models/Traits/HasMedia.php
```diff
@@ -88,7 +88,7 @@ public function addMedia(UploadedFile $file, string $collection = 'default', arr
         $filename = Str::uuid().'.'.$normalizedExt;
         $path = 'medias/'.$filename;
 
-        Storage::put($path, $normalizedBytes);
+        Storage::disk('public')->put($path, $normalizedBytes);
```
paulocastellano: @albertvisuals Same here

app/Models/Traits/HasMedia.php
```diff
@@ -126,7 +126,7 @@ public function addMediaFromPath(string $filePath, string $originalFilename, str
         $filename = Str::uuid().'.'.$normalizedExt;
         $storagePath = 'medias/'.$filename;
 
-        Storage::put($storagePath, $normalizedBytes);
+        Storage::disk('public')->put($storagePath, $normalizedBytes);
```
paulocastellano: @albertvisuals Same here

paulocastellano (changes requested):

database/migrations/2026_01_14_232321_create_posts_table.php
```diff
@@ -18,7 +18,7 @@ public function up(): void
             $table->uuid('workspace_id');
             $table->uuid('user_id')->nullable();
             $table->text('content')->nullable();
-            $table->json('media')->default('[]');
+            $table->json('media')->nullable()->default(null);
```
paulocastellano, line 21: It's good!

database/migrations/2026_04_15_000001_create_post_comments_table.php
```diff
@@ -16,7 +16,7 @@ public function up(): void
             $table->uuid('user_id')->nullable();
             $table->uuid('parent_id')->nullable();
             $table->text('body');
-            $table->json('reactions')->default('[]');
+            $table->json('reactions')->nullable()->default(null);
```
paulocastellano, line 19: It's good!

paulocastellano (approved):

app/Http/Controllers/App/AssetController.php
sebastian-works: Understood the concern about flexibility. However, the issue is that Storage::put() without a disk specification does not respect FILESYSTEM_DISK — we tested both FILESYSTEM_DISK=local and FILESYSTEM_DISK=public and in both cases files were written to storage/app/private, making them inaccessible via the web. The Storage::disk('public') was the quick fix to make local file serving work correctly.

paulocastellano: Probably you missed to ran:

```
php artisan storage:link
```

Required to `public` disk.

Can you remove it? So i can merge the changes related to MySQL.

sebastian-works: I did run php artisan storage:link — the symlink was confirmed correct. The issue was that Storage::put() was writing to storage/app/private regardless of FILESYSTEM_DISK=public. Happy to revert if you can confirm this works correctly on your end with FILESYSTEM_DISK=public. We can also split this into a separate PR for the storage fix and keep this PR focused on the MySQL migrations only.

paulocastellano: @albertvisuals sure, let's split into two PR's:

* one for mysql
* one for storage

paulocastellano: @albertvisuals i tested with:

```
FILESYSTEM_DISK=public
```

After i ran:

```
php artisan storage:link
```

It's working fine.

Maybe you need ran:
```
php artisan config:clear
````

Comment (sebastian-works) [reply app/Http/Controllers/App/AssetController.php]: Thanks yeah. missing php artisan config:clear was the culprit.

Comment (paulocastellano) [reply app/Http/Controllers/App/AssetController.php]: Also i have updated the docs:

https://docs.trypost.it/self-hosting/configuration#file-storage

Can you update the PR?

File: app/Http/Controllers/App/Settings/ProfileController.php [hunk 6]
```diff
@@ -126,8 +126,8 @@ public function destroy(ProfileDeleteRequest $request): RedirectResponse
 
             $user->workspaces()->detach();
 
-            if ($account) {
-                $account->delete();
+            if ($account && $user->isAccountOwner()) {
```
paulocastellano: @albertvisuals i'm fixing that in another PR, can you remove it?

paulocastellano (changes requested):

app/Http/Controllers/App/AssetController.php
sebastian-works: I have reverted the storage changes, they should be already gone from the PR

app/Http/Controllers/App/Settings/ProfileController.php
sebastian-works: I have removed it

PR #51, opened 2026-05-19
feat(auth): self-hosted registration gate + admin seeder

### LINKED ISSUE #46: [Feature Request] REGISTRATION_ENABLED env variable to disable open sign-ups for self-hosted installs
## Problem

When self-hosting TryPost, there is currently no way to disable open registration. This means anyone who finds the URL can sign up and create workspaces on your private install — which is a real concern for teams and organizations running TryPost internally.

## Proposed Solution

Add a `REGISTRATION_ENABLED` environment variable that controls whether open sign-ups are allowed:

```env
REGISTRATION_ENABLED=false
```

When set to `false`:
- The sign-up page should reject new registrations
- The "Sign up" link on the login page should be hidden
- **Workspace invites should still work** — invited users must be able to complete their signup via an invite link even when open registration is disabled

## Why This Matters

Self-hosted installs are typically private — used by a single organization or small team. Open registration defeats the purpose of running your own instance. An invite-only mode is standard practice for self-hosted tools (n8n, Gitea, Plausible, etc.).

Thanks for building a great open-source tool!

paulocastellano:
Closes #46.

## Summary

Self-hosted installs (the `SELF_HOSTED=true` default) now close `/register` to the public — your instance only exists for you and the people you invite.

- New `EnsureRegistrationEnabled` middleware gates `GET` and `POST /register`. Allowed when (a) `SELF_HOSTED=false`, (b) the URL carries \`?invite={id}\`, or (c) the session already has \`pending_invite_id\`. The query-param path persists into the session so the follow-up `POST` (which has no query string) still passes through.
- `AcceptInvite.vue` adds \`invite\` to the register link's query so invitees land on a working signup page even in self-hosted mode.
- `Login.vue` hides the "Sign up" link when self-hosted.
- `RegisteredUserController::store` clears \`pending_invite_id\` after signup.

## New seeder

`UserSeeder` bootstraps a single admin so a fresh self-hosted install can actually log in:

| | |
|---|---|
| Email | `admin@trypost.it` |
| Password | `password` |

Idempotent: skips if any user already exists. **Not** included in `DatabaseSeeder` — operator runs it once with \`php artisan db:seed --class=UserSeeder\`. Documented in the docs repo (`self-hosting/installation.mdx` step 3).

## OAuth signup (out of scope)

The Google/GitHub callback also creates accounts. It is **not** gated by this middleware. Wiring those callbacks into the same gate is a separate follow-up.

## Test plan

- [x] `php artisan test --compact --parallel` — 1592/1592 passing (baseline 1583 + 9 new)
- [x] Cases covered for `/register` and `AcceptInvite::show`:
  - `SELF_HOSTED=false` open signup (existing tests, now explicit)
  - `SELF_HOSTED=true` + no invite → 404 (`GET` and `POST`)
  - `SELF_HOSTED=true` + `?invite=…` → renders, persists to session
  - `SELF_HOSTED=true` + session has `pending_invite_id` → renders
  - `SELF_HOSTED=false` + invite in session → still works
  - Signup clears `pending_invite_id` from session
  - `AcceptInvite::show` renders for guest in both modes
  - Google/GitHub auth toggle props correctly delivered when self-hosted via invite
- [ ] Manual: full invitation round-trip on a `SELF_HOSTED=true` instance.

### DIFF
--- a/app/Http/Controllers/Auth/RegisteredUserController.php
+++ b/app/Http/Controllers/Auth/RegisteredUserController.php
@@ -57,6 +57,8 @@ class RegisteredUserController extends Controller
 
         Auth::login($user);
 
+        $request->session()->forget('pending_invite_id');
+
         if ($redirect = $request->input('redirect')) {
             if (str_starts_with($redirect, '/') && ! str_starts_with($redirect, '//')) {
                 return redirect($redirect);
--- a/app/Http/Middleware/App/EnsureRegistrationEnabled.php
+++ b/app/Http/Middleware/App/EnsureRegistrationEnabled.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Http\Middleware\App;
+
+use Closure;
+use Illuminate\Http\Request;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+class EnsureRegistrationEnabled
+{
+    public function handle(Request $request, Closure $next): mixed
+    {
+        if (! config('trypost.self_hosted')) {
+            return $next($request);
+        }
+
+        if ($inviteId = $request->query('invite') ?? $request->session()->get('pending_invite_id')) {
+            $request->session()->put('pending_invite_id', $inviteId);
+
+            return $next($request);
+        }
+
+        throw new NotFoundHttpException;
+    }
+}
--- a/bootstrap/app.php
+++ b/bootstrap/app.php
@@ -1,6 +1,7 @@
 <?php
 
 use App\Http\Middleware\Api\LoadWorkspaceFromToken;
+use App\Http\Middleware\App\EnsureRegistrationEnabled;
 use App\Http\Middleware\App\HandleInertiaRequests;
 use App\Http\Middleware\App\SetLocale;
 use Illuminate\Foundation\Application;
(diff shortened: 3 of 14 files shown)


sebastian-works:
Could we add a GOOGLE_LOGIN_ENABLED flag so that we can activate the YouTube integration separately from the Google Login capability. Or can we simply add SOCIAL_LOGIN_ENABLED to deactivate the social login options?

paulocastellano:
> Could we add a GOOGLE_LOGIN_ENABLED flag so that we can activate the YouTube integration separately from the Google Login capability. Or can we simply add SOCIAL_LOGIN_ENABLED to deactivate the social login options?

@albertvisuals we already have it:

To disable only social auth, use: `GOOGLE_AUTH_ENABLED=false`

To disable channels, use: `YOUTUBE_ENABLED=false`

sebastian-works:
Ah... thanks! Claude was pointing me wrongly there. 
Splendid and many thanks for the fast fixes

sebastian-works:
Is it possible that some of the changes block invites now?
I have just invited two people, but when they click Setup Account after The Accept Invite, they get the 404 NOT Found error in a pop-up window,  instead of the account setup screen.

paulocastellano:
@sebastian-works can share a video please?

PR #93, opened 2026-06-11
Fix: resolve Bluesky @mention handles to DIDs (mentions caused "Invalid post data")

Schrall:
## Problem

Any scheduled Bluesky post that contains an `@mention` fails to publish with **`Invalid post data`** (AT Protocol `InvalidRequest`). Posts without a mention publish fine.

### Root cause

In `BlueskyPublisher::parseFacets()`, the mention facet is built with the raw **handle** in the `did` field:

```php
'$type' => 'app.bsky.richtext.facet#mention',
'did' => $handle, // Will be resolved by Bluesky
```

That comment is incorrect. An `app.bsky.richtext.facet#mention` requires the target's **DID** (e.g. `did:plc:…`) — Bluesky does **not** resolve a handle placed in `did`. Sending a handle there makes the whole `app.bsky.feed.post` record invalid, so `com.atproto.repo.createRecord` rejects it with `InvalidRequest`, surfaced to the user as "Invalid post data". The result is that *every* post containing a mention silently fails.

## Fix

Resolve each handle to a DID via `com.atproto.identity.resolveHandle` before building the facet:

- Tries the account's own PDS first (authenticated), then falls back to the public AppView (`public.api.bsky.app`) and `bsky.social`.
- Resolutions are cached per post (a handle mentioned twice is resolved once).
- If a handle **cannot** be resolved, the mention facet is skipped so the post still publishes with the `@handle` as plain text, rather than failing the entire post.

This keeps mentions working as real, clickable, notifying mentions and removes the failure mode entirely.

## Tests

Adds two feature tests to `tests/Feature/Services/Social/BlueskyPublisherTest.php`:

- `bluesky publisher resolves mentions to DIDs as facets` — asserts the mention facet carries the resolved DID, not the handle.
- `bluesky publisher skips mention facet when handle cannot be resolved` — asserts graceful degradation (post still publishes, handle stays as plain text, no mention facet).

## Notes

- No new dependencies; uses the existing `socialHttp()` client.
- Verified end-to-end against a live self-hosted instance: before the fix every mention post returned "Invalid post data"; after it, the published record contains a proper `#mention` facet with the resolved `did:plc:…`.


paulocastellano:
@Schrall i will review that soon.

PR #110, opened 2026-06-23
feat(bluesky): support video upload and embedding

Schrall:
## Problem

`BlueskyPublisher` only attaches **images** — it iterates media and calls `uploadBlob()` exclusively for `isImage()` items. When a scheduled post's media is a **video**, the media is silently skipped and the post publishes as **text-only with no embed**. The post platform still reports `published`, so the failure is invisible until you look at the live feed.

Repro: schedule a Bluesky post whose only media is an `.mp4` → it goes out with no video.

## Why a separate path is needed

Bluesky videos are **not** uploaded to the PDS via `com.atproto.repo.uploadBlob` like images. They go through the dedicated **video service** (`video.bsky.app`), which transcodes the file (HLS) and writes the resulting blob back to the account's PDS. Only then can the blob be embedded as `app.bsky.embed.video`.

## What this PR does

Adds a video branch in `publish()` that runs only when no image embed was built (a post carries either images or a single video, never both — mirroring the official client):

1. **Resolve the real PDS host** from the account's DID document (`plc.directory` for `did:plc`, `/.well-known/did.json` for `did:web`). Entryway accounts store `https://bsky.social` as `meta.service`, but the video service-auth audience must be the account's actual `*.host.bsky.network` PDS.
2. **Mint a service-auth token** (`com.atproto.server.getServiceAuth`, `lxm=com.atproto.repo.uploadBlob`, `aud=did:web:<pds-host>`, 30‑min exp).
3. **Upload** the bytes to `app.bsky.video.uploadVideo` (handles the `409 already_exists` duplicate case by reusing the finished job's blob).
4. **Poll** `app.bsky.video.getJobStatus` until `JOB_STATE_COMPLETED`, then embed the blob as `app.bsky.embed.video`.

Failure modes (download error, oversized > 100 MB, service-auth failure, `JOB_STATE_FAILED`, timeout) are logged and return `null`, so the post still publishes as text instead of crashing the queued job — consistent with how `uploadBlob()` already degrades for images.

New NSID constants live in `BlueskyLexicon`; the video service host, its DID, and the PLC directory are configurable via `config/trypost.php` (env‑overridable).

## Tests

Adds three feature tests to `BlueskyPublisherTest`:
- uploads a video and embeds it as `app.bsky.embed.video`, asserting it goes to the video service (never `uploadBlob`);
- the upload service-auth audience is scoped to the **resolved PDS host**, not the entryway;
- a `JOB_STATE_FAILED` transcode degrades to a text-only post rather than throwing.

Full `BlueskyPublisherTest` suite (22 tests) and `pint` both pass locally.

## Notes / limitations

- `aspectRatio` is intentionally omitted (optional in the lexicon) since dimensions aren't reliably available server-side without ffprobe. Videos still render; this can be added later.
- No `getUploadLimits` preflight; an unverified-email / quota-exhausted account surfaces via the upload/job error path.

### DIFF
--- a/app/Enums/PostPlatform/ContentType.php
+++ b/app/Enums/PostPlatform/ContentType.php
@@ -217,6 +217,20 @@ enum ContentType: string
     }
 
     /**
+     * Whether a single post may carry images and a video together. Most
+     * targets accept only one or the other; this is conservative and lists
+     * only the types we've confirmed reject mixed media (e.g. a Bluesky post
+     * embed is `images` XOR `video`, never both).
+     */
+    public function supportsMixedMedia(): bool
+    {
+        return match ($this) {
+            self::BlueskyPost => false,
+            default => true,
+        };
+    }
+
+    /**
      * Whether this content type carries a text caption visible to viewers.
      * Stories are image-overlay only — viewers don't see a separate caption.
      */
--- a/app/Rules/ContentTypeCompatibleWithMedia.php
+++ b/app/Rules/ContentTypeCompatibleWithMedia.php
@@ -61,6 +61,10 @@ class ContentTypeCompatibleWithMedia implements DataAwareRule, ValidationRule
         if ($hasVideo && ! $contentType->supportsVideo()) {
             $fail("{$contentType->label()} does not support videos.");
         }
+
+        if ($hasImage && $hasVideo && ! $contentType->supportsMixedMedia()) {
+            $fail("{$contentType->label()} can't combine an image and a video in the same post.");
+        }
     }
 
     /**
--- a/app/Services/Social/BlueskyLexicon.php
+++ b/app/Services/Social/BlueskyLexicon.php
@@ -27,8 +27,6 @@ final class BlueskyLexicon
 
     public const VIDEO_GET_JOB_STATUS = 'app.bsky.video.getJobStatus';
 
-    public const VIDEO_GET_UPLOAD_LIMITS = 'app.bsky.video.getUploadLimits';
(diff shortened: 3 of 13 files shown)


Schrall:
Thanks for the review — all five points addressed in `7f648f8`:

1. **`tempnam()` may return `false`** (L208) — now guarded; bail with a log before entering the try block.
2. **`@unlink($tempFile)` TypeError on non-string** (L301) — `finally` now guards with `is_string()`.
3. **`fopen()` may return `false`** (L255) — checked in the new `attemptVideoUpload()` helper; returns `null` if the file can't be opened.
4. **`did:web` with path segments** (L401) — now maps colon segments to host + path per the did:web spec (`did:web:example.com:user:alice` → `https://example.com/user/alice/did.json`); a bare host still uses `/.well-known/did.json`.
5. **`pollVideoJob()` ignored non-2xx `getJobStatus`** (L326) — now bails early with a log on `$response->failed()` instead of sleeping to the timeout.

Additionally, while validating against the live video service I hit a **transient** `JOB_STATE_FAILED "Failed to process video"` for otherwise-valid input. The upload+poll is now retried up to `VIDEO_UPLOAD_ATTEMPTS` (3) times — extracted into `attemptVideoUpload()`, with the upload service-auth token minted once and reused across attempts. Added a test covering retry-after-transient-failure; full `BlueskyPublisherTest` suite (23) is green.

paulocastellano:
@Schrall i will review it, thank you for submit!

PR #122, opened 2026-06-28
fix: make external-URL media safe end-to-end (API hosting, X MIME, error sanitization)

paulocastellano:
All of this comes from one incident: a customer's posts (created via the public API with bare external-URL media — an `images.weserv.nl` proxy URL stored verbatim, no `mime_type`/`path`) failed to publish across Facebook, X, and Instagram, and the X failure leaked a server file path into the failure email.

## 1. API downloads & hosts external media (root cause)

The public REST API accepted inline post `media` as a free-form array and stored it verbatim, so a post could reference a third-party URL we never hosted. Publishing then depended on that URL staying alive — when the proxy 404'd, every platform failed.

Inline media URLs on create/update now go through the same **download → MIME-validate → host** path as the attach-from-url endpoint (`MediaAttacher`). External URLs become hosted media on our storage; already-hosted items (carrying a `path`) pass through. If any URL can't be fetched the request is rejected with **422** and nothing is persisted. MCP and the web flow were already safe (MCP takes no inline media; web uploads are pre-hosted).

## 2. X recovers a missing MIME (defense)

A bare-URL item has no `mime_type` and no `path` to infer one from, so `XPublisher::getMediaCategory()` (param typed `string`) received `null` → **TypeError**. X downloads the bytes anyway, so it now recovers the MIME from the downloaded file and fails with a clean exception if it still can't. `chunkedUpload` also takes `?string $mediaCategory` now (and only sends `media_category` when present) so the recovered-but-uncategorizable path can't TypeError. Audited every publisher — X was the only one whose strict-mime path was reachable with null.

## 3. Failure emails never leak raw errors (security)

The post-failure email renders each platform's `error_message` verbatim. The TypeError's message embeds the server path (`.../releases/<id>/app/Services/Social/XPublisher.php on line 130`), which the publish job's catch-all (and the job-failed hook) stored raw → it reached the customer's inbox. Now only our own `SocialPublishException` (vetted `userMessage`) is surfaced; any other throwable — engine errors like `TypeError`, or library exceptions such as Intervention's decoder errors which embed temp paths — is replaced with a generic line, with the raw detail kept in logs. Token-expired and reschedule paths are unchanged.

## 4. Deliberate publish exceptions are typed (UX, follows from #3)

With #3 genericizing anything that isn't a `SocialPublishException`, the deliberate user-facing `\Exception` throws in the X / Facebook / Threads / LinkedIn (+Page) publishers would have been genericized too — losing actionable messages like "Unsupported media type for Facebook". Those are now their platform's `SocialPublishException` with a vetted `userMessage` + category, so failures keep a specific, safe message and the generic line is reserved for genuinely unexpected errors. (Left as plain `\Exception` on purpose: the download helpers in Bluesky/Mastodon — caught internally, never reach the email — and the shared content-length guard in `HasSocialHttpClient`.)

## 5. Centralized inline-media validation

The `media.*` validation rules were duplicated across the web update request and both API requests, and had diverged (web requires hosted `id`+`path` and tracks `source`; the API accepts a bare external `url`). They're now one source of truth — `App\Support\PostMediaRules::rules(hosted:)`, same pattern as `PostPlatformMetaRules` — parameterized by contract, so there's a single place to add a media key and the `validated()`-strips-unlisted-keys footgun can't drift between entry points. Behavior is reproduced exactly; the web store's loose `media => array` is intentionally left as-is.

## Scope note
The fix resolves the three errors **for newly created/updated posts**. An already-stored broken post published via `PUT` with only `status` (no `media`) skips hosting and still fails — remediation for the customer's existing posts: re-submit the post with its `media` (re-hosts it), then publish.

## Tests
- API: external URL hosted on create/update; unreachable URL → 422, nothing persisted; already-hosted item passes through with zero downloads.
- X: missing MIME recovered from bytes → publishes (no TypeError); clean `XPublishException` when media can't be downloaded.
- Publish job: raw internal errors (incl. a `TypeError` with a path) are genericized in both the catch-all and the `failed()` hook, and the path never reaches the record; vetted publish-exception messages preserved.
- `PostMediaRules`: both contract variants locked (hosted vs external) + shared keys preserved.
- Full suite: **2341 passing**.


rabucho:
Thanks, I can confirm the fix works.

PR #129, opened 2026-07-01
feat(sidebar): make nav groups collapsible; workspace group starts closed

dantaspaulo:
## Problem

The sidebar renders every nav group expanded. With the main, posts, workspace
and support groups all open at once, the secondary "Workspace" group (channels,
signatures, labels, assets) competes with the primary flow for attention.

## Change

`NavMain` gains two optional props:

- `collapsible` (default `false`) — when set, the group label becomes a
  `Collapsible` trigger with a rotating chevron.
- `defaultOpen` (default `true`) — the initial open state when collapsible.

When `collapsible` is not set, the group renders exactly as before (always
open), so every other caller is unaffected.

`AppSidebar` then marks only the **Workspace** group as
`collapsible :default-open="false"`, so it starts collapsed. The main, posts and
support groups are unchanged.


paulocastellano:
We need improve it, only for workspace don't make sense.

PR #131, opened 2026-07-01
feat(profile): crop avatar/logo before upload (round mask by default)

dantaspaulo:
## Problem

Avatar/logo upload is raw: any aspect ratio goes straight to the server, so the
circular avatar ends up off-center. Since the avatar/logo also renders as a
circle on generated cards, letting the user frame it improves the result a lot.

## Change

Selecting a file now opens an `ImageCropperDialog` (zoom + drag) with a **round
mask by default**. Save produces a `canvas.toBlob` → `File` that is uploaded
already cropped and resized (512×512 output, well under the 2 MB limit).

The integration lives in one place — `PhotoUpload.vue`, used by both the profile
avatar and the workspace logo. The cropper mounts only after the dialog has its
final layout, otherwise the circular stencil degrades into a square inside the
modal.

- `ImageCropperDialog.vue` (new) — the crop modal (`CircleStencil`, `RectangleStencil` for a square shape).
- `PhotoUpload.vue` — opens the crop before uploading.
- `lang/{en,es,pt-BR}/common.php` — `photo_upload.crop_*` strings.

## New dependency

This adds **`vue-advanced-cropper` (^2.8.9)**. It's a mature, Vue 3-compatible
cropper; the `package-lock.json` change is limited to that subtree
(`classnames`, `debounce`, `easy-bem`, `vue-advanced-cropper`). Happy to swap it
for another cropper if you'd prefer a different one.

### DIFF
--- a/lang/en/common.php
+++ b/lang/en/common.php
@@ -19,6 +19,11 @@ return [
         'uploading' => 'Uploading...',
         'remove' => 'Remove photo',
         'hint' => 'Recommended: square image, max 2 MB.',
+        'crop_title' => 'Crop image',
+        'crop_description' => 'Drag and zoom to frame it. The area inside the circle will be used.',
+        'crop_hint' => 'Drag to position',
+        'crop_save' => 'Save',
+        'crop_cancel' => 'Cancel',
     ],
 
     'timezone' => [
--- a/lang/es/common.php
+++ b/lang/es/common.php
@@ -19,6 +19,11 @@ return [
         'uploading' => 'Subiendo...',
         'remove' => 'Eliminar foto',
         'hint' => 'Recomendado: imagen cuadrada, máximo 2 MB.',
+        'crop_title' => 'Recortar imagen',
+        'crop_description' => 'Arrastra y usa el zoom para encuadrar. Se usará el área dentro del círculo.',
+        'crop_hint' => 'Arrastra para posicionar',
+        'crop_save' => 'Guardar',
+        'crop_cancel' => 'Cancelar',
     ],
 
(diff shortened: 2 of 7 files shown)


paulocastellano:
Thanks a lot for this, @dantaspaulo — the UX idea is spot on. Cropping the avatar/logo before upload, with the mask matching how it actually renders, genuinely improves the result, and we want it in the product.

We're going to close this one in favor of a from-scratch implementation, for two reasons:

1. **The dependency.** `vue-advanced-cropper` is effectively unmaintained — the last release (2.8.9) was June 2024, ~2 years ago. We'd rather not take on an abandoned package for something this load-bearing.

2. **The crop shape.** Avatars and workspace logos don't render the same way (one's a circle, the other a rounded square), so the crop mask should follow the display shape instead of always being round.

Since what we need is narrow — fixed 1:1, a circle/square mask, fixed-size output — we're rebuilding it dependency-free with a small canvas-based cropper, which also lets us cover it with tests.

None of this takes away from the contribution: the idea and the integration points (the `PhotoUpload` flow, the i18n strings) were a great starting point and directly shaped the replacement. Really appreciate you taking the time — please keep them coming. 🙏

dantaspaulo:
That's great, @paulocastellano. I've taken your suggestion on board; I'm going to remove the current version from my end and implement it the way you described.

paulocastellano:
> That's great, @paulocastellano. I've taken your suggestion on board; I'm going to remove the current version from my end and implement it the way you described.

I already implemented it:

https://github.com/trypostit/trypost/pull/174

I will merge soon.

PR #144, opened 2026-07-01
fix(media): don't 500 when the Unsplash/Giphy key is set to null

dantaspaulo:
## Problem

`config('services.unsplash.access_key', '')` only falls back to `''` when the
key is **absent**. If the env var is present but empty, `config()` returns
`null`, and assigning `null` to the `string` property throws a `TypeError` in the
constructor — taking the whole request down with a 500 during media search.
`GiphyService` has the same issue with `services.giphy.api_key`.

## Fix

Cast to `(string)` so a null/empty key degrades gracefully to `''`. Both services
already return an empty result set when the key is empty (`if (empty(...))`), so
this restores that intended behavior instead of throwing.

## Tests

`MediaSearchKeyFallbackTest` — constructing each service with a `null` key and
calling `search()` returns an empty result set without throwing.

### DIFF
--- a/app/Services/GiphyService.php
+++ b/app/Services/GiphyService.php
@@ -15,7 +15,7 @@ class GiphyService
 
     public function __construct()
     {
-        $this->apiKey = config('services.giphy.api_key', '');
+        $this->apiKey = (string) config('services.giphy.api_key', '');
     }
 
     /**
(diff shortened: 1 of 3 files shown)


paulocastellano:
Thanks @dantaspaulo!

PR #146, opened 2026-07-01
fix(workspace): default content_language to the app locale, not 'en'

dantaspaulo:
## Problem

When a workspace is created without an explicit `content_language`,
`data_get($data, 'content_language')` returns `null`, `array_filter` drops the
key, and the column falls back to its DB default of `'en'`. So AI content and
notifications came out in English even when the app was running in another
locale.

## Fix

Default to `app()->getLocale()`:

```php
'content_language' => data_get($data, 'content_language', app()->getLocale()),
```

Since `getLocale()` is never null, `array_filter` keeps the key and a new
workspace inherits the user's current language. An explicit `content_language`
still wins.

## Tests

`CreateWorkspaceTest` — a workspace with no `content_language` inherits the app
locale (`pt-BR`), and an explicit value (`es`) is kept over the locale.


paulocastellano:
@dantaspaulo thank you!

PR #179, opened 2026-07-11
Add per-image alt text for accessibility across 9 platforms

paulocastellano:
## What

Lets users add **alt text** (image accessibility descriptions) to their media, published to every platform whose API supports it. A customer asked for this specifically for Bluesky; this covers all supported platforms in one go.

One alt text per image, entered once in the composer, sent to each platform that accepts it. Alt text is entirely **optional** — nothing requires it.

## Platform support

Researched against each platform's official API docs:

| Supported (9) | Field | Cap |
|---|---|---|
| Bluesky | `alt` (required on embed) | 2000 |
| X | `alt_text.text` (v2 `media/metadata`) | 1000 |
| Mastodon | `description` (media upload) | 1500 |
| LinkedIn | `altText` | 4086 |
| Facebook | `alt_text_custom` | 1000* |
| Instagram | `alt_text` (images only) | 1000 |
| Threads | `alt_text` | 1000* |
| Pinterest | `alt_text` | 500 |
| Discord | `description` (per attachment) | 1024 |

*Facebook/Threads document no limit — defensive cap. **Not supported:** TikTok, Telegram, YouTube (no-op).

## How it's built

- **Storage:** per-image, in the existing `MediaItem.meta.alt_text` — **no migration**. Shared across all platforms of a post.
- **Single source of truth for caps:** `Platform::altTextMaxLength()` / `supportsAltText()`. No hardcoded caps in publishers; each truncates via the enum with `mb_substr`.
- **Validation:** `media.*.meta.alt_text` (`nullable`, `max:2000`) in `PostMediaRules` (web + API). The open `meta` array rule still passes width/height/duration through.
- **Publishers:** each supporting publisher reads `$media->altText()` and injects the platform's field at its upload point. Bluesky sends `''` when empty (its field is required); everyone else omits the field when there's no description.
- **Pinterest reconciliation:** migrated off an orphan per-platform `postPlatform.meta.alt_text` (read by the publisher, no UI, hardcoded cap) to the per-image alt like every other platform — one source of truth.
- **LinkedIn:** stopped sending `original_filename` as a junk `altText` fallback; now sends the real description or omits it.
- **Frontend:** a per-image "Alt" button + "ALT" badge in the composer (image items only), a small `AltTextDialog` with a character counter and accessibility hint, and the `alt_text` type. i18n in all 15 locales.

## Testing

- Per-publisher `Http::fake` tests asserting the alt appears in the outgoing payload (each verified to fail before its fix), plus truncation, null-omission, and image-only gating (Instagram/Threads don't send alt on videos).
- Validation tests (persist + >2000 rejection) and the Pinterest meta-ignored regression guard.
- Full backend suite green (2512 passed); Pint clean; localization parity green; eslint clean on the touched frontend files.

## Out of scope (v1)

- AI-generated alt text (manual entry only).
- Setting alt text via the MCP media-attach tools (the composer covers the need; a follow-up would add an optional `alt_text` to those two tools).


sayhitogambit:
Alt text é essencial, parabéns, release bem sólido!!

PR #193, opened 2026-07-20
feat(ai-create): let users rate the draft the AI produced

dantaspaulo:
> **Stacked on #191** — the review screen this control sits on comes from that PR. The first commit here is #191's; review only the second.

## Problem

There is no signal for whether an AI generation was actually any good. Publishing tells you nothing: a user can publish something mediocre, or abandon a great draft. So there is no way to tell whether a prompt change, a new template or a model swap made generations better or worse.

## Change

A small 1–5 star control under the review blocks. It is optional, never blocks anything, and answers `204` so the UI can fire and forget.

- A rating tied to a specific item is **idempotent** — rating again updates it rather than inserting a second row. That is enforced by a unique index, not just in the application. Standalone ratings stay possible, since NULLs are distinct in a unique index.
- `rateable_type` is validated against the registered morph aliases, so an arbitrary type cannot blow up later when the relation is resolved.
- The control follows the WAI-ARIA radiogroup pattern: arrows move focus and selection, Enter or click submits.

The table is polymorphic, so the same control can be dropped anywhere else you want a quality signal later.

## Tests

Five feature tests: recording, authentication, the 1–5 bound, the morph-map guard, and idempotency. The full suite passes (2643), Pint and ESLint are clean.

## Notes

- Translations follow the same approach as #191: `en` and `pt-BR` written properly, the other thirteen locales carrying the English strings as an untranslated fallback so the parity test passes.
- The original version of this also snapshotted a "brand completeness" score alongside each rating, to correlate output quality with how well the workspace brand was filled in. That depended on a helper that does not exist here, so I dropped it — the rating is just the rating. Happy to add something equivalent if you would find that correlation useful.

### DIFF
--- a/app/Broadcasting/UserAiDraftChannel.php
+++ b/app/Broadcasting/UserAiDraftChannel.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Broadcasting;
+
+use App\Models\User;
+
+class UserAiDraftChannel
+{
+    public function join(User $user, User $owner, string $draftId): bool
+    {
+        return $user->is($owner);
+    }
+}
--- a/app/Enums/Ai/DraftStatus.php
+++ b/app/Enums/Ai/DraftStatus.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Enums\Ai;
+
+/**
+ * Lifecycle of an AI post draft: the two-phase "review before final generation"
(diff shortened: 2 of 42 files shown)


paulocastellano:
@dantaspaulo can you attach a video demo about this please?

dantaspaulo:
@paulocastellano seria uma etapa previa a geração do conteúdo, de modo que o usuário consiga revisar e editar antes.

https://github.com/user-attachments/assets/c72a0a1e-ed23-4958-8fcf-dc8bea383859

PR #198, opened 2026-07-23
fix: allow unicode filenames in chunked asset uploads

paulocastellano:
## Summary
- Chunked uploads put the filename in the `X-File-Name` header; browsers reject non–ISO-8859-1 bytes (e.g. en-dash `–` in `Corte 6 – Quantidade…‥.mp4`), so `fetch` fails client-side and the user only sees a generic toast.
- Percent-encode the name in `chunkedUpload.ts` and `rawurldecode` it in `StoreChunkedAssetRequest` before validation/storage.
- Feature tests cover en-dash, emoji, spaces/`..ext`, plain ASCII (backwards compatible), and still-rejected extensions.

## Test plan
- [x] `php artisan test --compact tests/Feature/ChunkedUploadFilenameEncodingTest.php tests/Feature/AssetControllerTest.php --filter=chunked`
- [x] Manually upload a file named like `Corte 6 – Quantidade ou qualidade_ Os dois..mp4` on Assets → Meus uploads
- [x] Confirm a simple ASCII name (e.g. `video.mp4`) still uploads

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/Http/Controllers/App/AssetController.php
+++ b/app/Http/Controllers/App/AssetController.php
@@ -120,7 +120,7 @@ class AssetController extends Controller
         };
 
         $filename = Str::uuid().'.'.$extension;
-        $path = 'medias/'.$filename;
+        $path = "medias/{$filename}";
 
         Storage::put($path, $response->body());
 
--- a/app/Http/Controllers/App/AssetController.php
+++ b/app/Http/Controllers/App/AssetController.php
@@ -9,6 +9,7 @@ use App\Http\Requests\App\Asset\StoreAssetRequest;
 use App\Http\Requests\App\Asset\StoreChunkedAssetRequest;
 use App\Http\Resources\App\MediaResource;
 use App\Models\Media;
+use App\Models\Workspace;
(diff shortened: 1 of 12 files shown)


paulocastellano:
Follow-up: large videos were still dying/hanging on the **last** chunk because we reassembled the file and uploaded the whole thing to R2 in that request (`set_time_limit(0)` only masked the 30s fatal — hence the 53s 200).

Videos/PDFs on S3 disks now use **multipart upload**: each chunk is an UploadPart, the last request only completes the multipart. Intermediate requests stay ~same size/time as before.

paulocastellano:
## Review follow-up

Went through the whole PR and fixed the real issues:

- **Orphan objects** — if Media create fails after multipart complete, delete the R2 key
- **Part ordering** — reject unexpected offsets; idempotent replay of already-accepted chunks
- **5 MiB rule** — reject undersized non-final parts (S3/R2 requirement)
- **Size metadata** — persist bytes actually received, not the client `total_size` header
- **Readability** — controller split into multipart vs local assemble + shared response helper

86 related tests green.

PR #216, opened 2026-07-31
add openrouter to eligible ai providers

rallisf1:
I don't get why you need this block of code everywhere, but I needed a quick fix to use openrouter, so here you go.

IMHO The best fix would be to get rid of this code block in all files and do a sanity check on `config('ai.default')` during bootstrap and show a proper warning if no/wrong provider is set.


paulocastellano:
@rallisf1 i think on new version of laravel/ai it's supporting open router.

Have you checked it ?

rallisf1:
You are right, I only checked v13 which is the stable. On the master branch they dropped all OpenAI compatible providers and say to use `OPENAI_BASE_URL` for them with `OPENAI_API_KEY`.

rallisf1:
closing in favor or #220

PR #220, opened 2026-08-01
Fix OpenRouter via laravel/ai default provider (revises #216)

paulocastellano:
Supersedes / revises #216.

## Why

#216 added `'openrouter' => Lab::OpenRouter` to every agent `provider()` match. That works on the surface, but:

1. **laravel/ai already supports OpenRouter** as a first-class driver. When an agent omits `provider()`, `Promptable` falls back to `config('ai.default')`.
2. The duplicated matches were the real bug: anything other than `openai`/`anthropic` was forced to **Gemini**, so `AI_TEXT_PROVIDER=openrouter` never reached OpenRouter.
3. #216's `BrandAnalyzerRunner` check used `config('services.openrouter.api_key')`, which **does not exist** — so brand autofill would still treat OpenRouter as unavailable.
4. The installed `laravel/ai` (`v0.5.1`) only supported OpenRouter for text/embeddings via the legacy Prism gateway — no native image/audio/STT support, no OpenRouter-specific credit-exhaustion error handling.

As noted in #216, the better fix is to stop hardcoding provider maps in every agent — and to actually get current on `laravel/ai` while we're in here, since the OpenRouter support keeps improving release to release.

## What changed

- Removed `provider()` from all 6 AI agents so they use `config('ai.default')` (supports `openrouter`, `openai`, `anthropic`, `gemini`, etc.).
- **Bumped `laravel/ai` `v0.5.1` → `v0.10.3`.** OpenRouter moved onto a native gateway (image generation, TTS/STT, web search, `InsufficientCreditsException`), `prism-php/prism` is no longer a dependency at all.
- Removed `default_text_model` / `AI_TEXT_MODEL` from `config/ai.php`. Every other modality (image, audio, transcription, embeddings, reranking) only ever picked a **provider** and let that provider use its own default model — text was the only one with a config-pinned model forced on top, which is exactly what broke OpenRouter: switching `AI_TEXT_PROVIDER` kept sending OpenAI's model id (`gpt-5.4`) to whichever provider actually ran. Removed `model()` from all 6 agents too, so `laravel/ai` resolves the model from whichever provider is active, same as image/audio/etc already do.
- Fixed `AiImageClient` hardcoding `gpt-image-2` regardless of `AI_IMAGE_PROVIDER` — same bug, image side. It now lets the SDK pick the model for whichever provider is configured, and reads the real provider/model back off the response instead of assuming one.
- Every `RecordAiUsage::recordText()`/`recordImage()` call site now records the provider/model the response actually reports (`$response->meta->provider` / `->model`) instead of assuming `config('ai.default')` — usage/credit logs were previously mislabeled whenever a non-default provider ran.
- Removed the `App\Support\AiConfiguration` wrapper class — it was one line of logic (`filled(config("ai.providers.{$provider}.key"))`) used by two call sites, with no `laravel/ai`-native equivalent to lean on instead (confirmed `AiManager`/`Provider` expose no `isConfigured()`/`hasKey()` check — the package's model is try-then-catch, not pre-flight). Inlined directly into `HandleInertiaRequests` and `BrandAnalyzerRunner`.
- `data_get()` cleanup for array reads across every file touched by the above.

## ⚠️ Breaking change for self-hosted installs: `AI_TEXT_MODEL` is gone

If your `.env` sets `AI_TEXT_MODEL`, it no longer does anything — remove it. The text model is now chosen automatically from whichever provider `AI_TEXT_PROVIDER` points at (OpenAI's own default is still `gpt-5.4`, so nothing changes if you're on the default `AI_TEXT_PROVIDER=openai` and never overrode the model). This only affects you if you had `AI_TEXT_MODEL` set to something other than `gpt-5.4`, or if you're pointing `AI_TEXT_PROVIDER` at a non-OpenAI provider.

**Needs a call-out in the release notes when this ships**, so self-hosted users upgrading know to drop `AI_TEXT_MODEL` from their `.env`.

## Setup for OpenRouter

```env
AI_TEXT_PROVIDER=openrouter
OPENROUTER_API_KEY=sk-or-v1-...
```

(No `AI_TEXT_MODEL` needed — OpenRouter defaults to `anthropic/claude-sonnet-4.6` unless overridden per-provider in `config/ai.php`.)

## Testing checklist

- [ ] Set `AI_TEXT_PROVIDER=openrouter` + `OPENROUTER_API_KEY` in `.env`, remove any `AI_TEXT_MODEL` line, then restart Horizon (`php artisan horizon:terminate`) so the `ai` queue picks up the new env
- [x] Brand analysis: Settings → Brand (or a fresh workspace's creation flow) → paste a site URL → Autofill — exercises `BrandAnalyzer`
- [x] Full AI post creation: Posts → Create → AI tab (sparkles icon) → generate a post — exercises `PostContentGenerator` + `PostContentHumanizer`
- [x] Inline AI rewrite on an existing post's compose editor ("Generate with AI") — exercises the `PostContentStreamer` streaming path specifically (the trickiest fix: `StreamableAgentResponse` doesn't expose `meta` publicly, worked around via `then()`)
- [x] AI image regeneration on a post's AI-generated media — exercises `PostImageRegenerator` (text side, now via OpenRouter) + `AiImageClient`/`TemplateImageGenerator` (image side, still OpenAI-only per `AI_IMAGE_PROVIDER` — unaffected by the text provider switch)
- [x] Check the `ai_usage_logs` table after each generation — `provider`/`model` columns should show the real values (e.g. `openrouter` / `anthropic/claude-sonnet-4.6`), not stale/hardcoded ones
- [x] Switch back to `AI_TEXT_PROVIDER=openai` (restart Horizon again) and confirm nothing regressed — OpenAI's resolved model should still be `gpt-5.4`

Closes #216 once merged (or we can close #216 manually in favor of this).


paulocastellano:
@rallisf1 are you able to test this pull request with openrouter key please?

rallisf1:
@paulocastellano currently on vacation and the internet sucks. You've got mail though

paulocastellano:
@rallisf1 it's working fine, thank you man.

It's merged!

PR #224, opened 2026-08-03
fix(tiktok): send an empty JSON object to creator_info, not an empty array

Goo6i:
TikTokCreatorInfo::fetchFresh() passes an empty PHP array as the request body. Laravel serialises that to [], and TikTok's creator_info/query endpoint expects an object, so it rejects the request:

```json
{"error":{"code":"invalid_params","message":"The request parameter type is incorrect"}}
```

That is a 4xx, so $response->failed() is true and the method returns emptyPayload(). The composer then has no creator data at all, no max_video_post_duration_sec, and an empty privacy_level_options, which falls through to the static list in Platform::publishConfig(). The user is offered privacy options that may not match what their account actually allows.

Sending {} returns the real payload:

```json
{"data":{"creator_username":"...","privacy_level_options":["FOLLOWER_OF_CREATOR","MUTUAL_FOLLOW_FRIENDS","SELF_ONLY"],"duet_disabled":true,"stitch_disabled":true,"max_video_post_duration_sec":3600}}
```

Those options come from a private account, which is why PUBLIC_TO_EVERYONE is missing. 

Verified against a live TikTok account before and after the change. Added a test asserting the request body is {} and it fails on current main and passes with the fix.


paulocastellano:
@Goo6i thank you man, i will test it locally before merge!

paulocastellano:
@Goo6i thank you man, it really fixed our issue.

Approved and Merged!

Issue #255, opened 2026-08-08
SocialAccountObserver::notifyOnboarding() lazy-loads workspace — crashes when 2+ SocialAccount rows are batch-hydrated


paulocastellano:
## Problem

`App\Observers\SocialAccountObserver::notifyOnboarding()` reads `$socialAccount->workspace?->account` without `workspace` being eager-loaded on the triggering model.

This is latent almost everywhere because Laravel only flips `preventsLazyLoading` on a hydrated batch when `Builder::hydrate()` loads **more than one distinct row** in the same query (`Illuminate\Database\Eloquent\Builder::hydrate()`, gated by `count($items) > 1`). A single-account code path never trips it. `Model::shouldBeStrict(! $this->app->isProduction())` is set in `app/Providers/AppServiceProvider.php:217`, so:

- **Local/testing:** strict mode is on → any caller that batch-processes 2+ distinct `SocialAccount` rows and triggers a status-changing `update()` (which fires the observer) throws `LazyLoadingViolationException` immediately.
- **Production:** strict mode is off → the same code path degrades to a silent N+1 query instead of a crash.

## Where this was found

Discovered while building `App\Jobs\VerifyUpcomingPostConnections` (PR for the proactive upcoming-post connection check). That job intentionally groups at-risk `post_platforms` by `social_account_id` and can call `SocialAccount::markAsTokenExpired()` on multiple distinct accounts within a single run — exactly the "2+ distinct rows hydrated, one triggers the observer" shape that surfaces the bug.

**Fixed at the call site for that job only** (`app/Jobs/VerifyUpcomingPostConnections.php`, `atRiskPostPlatforms()`): eager-loads `socialAccount.workspace` before any account in the batch can be updated. That closes the gap for this one job, not for the observer itself.

## Why this needs a real fix, not another per-caller workaround

Any other current or future code path that batch-updates 2+ distinct `SocialAccount` rows to a different status without separately remembering to eager-load `workspace` will hit the same crash (locally/in tests) or silent N+1 (in production). Audited during the review of the job above — none of the other current call sites are exposed today (each operates on a single account per invocation), but that's incidental, not structural:

- [ ] `app/Jobs/RefreshSocialToken.php` — single account per job, not exposed today
- [ ] `app/Jobs/PublishToSocialPlatform.php` — single account per job, not exposed today
- [ ] `app/Http/Controllers/Auth/*Controller.php` (OAuth connect flows) — single account per request, not exposed today
[...]

PR #256, opened 2026-08-08
feat: proactive connection check for at-risk posts + SocialAccount name centralization

paulocastellano:
## Summary

This branch does two things — they ended up intertwined because the second grew out of a defensive-null-check discussion during review of the first.

### 1. Proactive connection check for posts about to publish

Closes a gap left after #254: the only proactive connection-health check (`social:check-connections`) runs once a day across every workspace, with no regard for what's about to publish. A token that dies right after that daily check runs can sit broken for up to 24h before anyone is told — including through a scheduled post's actual publish attempt failing with no advance warning.

This adds a third, narrower layer:

- **`social:check-upcoming-connections`** (new command, every 15 min) finds the distinct workspaces with at least one `post_platform` due to publish within the next hour, and dispatches `VerifyUpcomingPostConnections` once per workspace — never once per post.
- **`VerifyUpcomingPostConnections`** (new job) groups at-risk `post_platforms` by `social_account_id` and calls `ConnectionVerifier::verify()` **once per distinct account**, regardless of how many posts share it. Accounts already `TokenExpired`/`Disconnected` are never re-verified — straight to the notification using the existing `error_message`. A broken account is marked `TokenExpired` (`notify: false`) — this job never escalates straight to `Disconnected`; that stays the daily job's responsibility, so there's one consistent escalation lifecycle app-wide. A `last_verified_at` cache (40 min) keeps a healthy account from being re-verified on every 15-minute tick.
- A new `PostAtRisk` notification (email + in-app), grouped by account, listing every affected post's scheduled time — sent immediately, not batched into the next daily digest. Reuses the existing `account_disconnected` email preference; no new toggle.
- `post_platforms.connection_warning_sent_at` is the per-post dedup marker, time-boxed to 1 day so a rescheduled post re-enters the risk window instead of being silently skipped forever.
- A 5-minute "recently disconnected" grace period avoids a duplicate email when this job and another process (e.g. the daily sweep) discover the same broken token within a few minutes of each other.

### 2. SocialAccount display-name/handle centralization

Dozens of frontend files each hand-rolled their own `username || display_name` (or the reverse) fallback for "what name do we show for this account", with inconsistent null-handling. This adds two backend sources of truth on `SocialAccount`:

- `handle()` / `handleLabel` — username-wins fallback chain, ending in the platform's own label (e.g. "TikTok") as an absolute last resort.
- `accountDisplayName()` / `displayLabel` — display_name-wins version of the same chain.

Both are guaranteed non-empty, non-null strings — every connector's OAuth scopes always populate at least one of username/display_name (verified per-platform, e.g. TikTok's `user.info.profile` scope is unconditionally requested). ~30 frontend files were migrated to consume `display_label`/`handle_label` instead of hand-rolling the fallback, and their TypeScript types dropped the now-unnecessary `| null`.

Two smaller cleanups rode along:

- Avatar-initial extraction (`.charAt(0)` / `.charAt(0).toUpperCase()`, scattered across ~15 files with inconsistent casing) now goes through the existing `getInitials()` composable — also picks up first+last-word initials for multi-word names instead of always just one letter.
- The "Manage notifications" email footer link only makes sense on preference-driven emails (post-published, account-disconnected, etc.), not transactional ones (password reset, workspace invite). Split into two Maizzle components (`footer.html` / `footer-authenticated.html`) instead of a conditional prop.

## Fixes along the way

- **#255** (filed during this branch): found and worked around an unrelated `SocialAccountObserver::notifyOnboarding()` lazy-loading bug that only surfaces when 2+ distinct `SocialAccount` rows are hydrated and updated in one run — exactly this job's shape. Root cause is out of scope here; tracked separately.
- Several rounds of independent review caught and fixed: hardcoded-English email text bypassing the 16-locale translation infra, missing index for the new 15-minute query, a permanent (not time-boxed) dedup marker, null `social_account_id`/ownerless-workspace edge cases, a duplicate-email race between concurrent discovery of the same broken token, queue payload bloat, an account hard-deleted mid-run aborting the whole job instead of just that account, `PostAtRisk`'s subject/count disagreeing with the in-app notification title, a dead `v-if` guard left over from the display-name migration that made an "@handle" line always render (including duplicating the name for accounts with no real username), a preview fallback leaking a raw platform slug instead of a human label, `$postCount` counting `post_platforms` instead of distinct posts, and `ConnectionVerifier::verify()` silently treating a Telegram/Discord `false` return (bot removed from channel/guild) as healthy instead of routing it through the same broken-account handling as every other platform.

## Test plan

- [x] Full suite: `php artisan test --compact --parallel` — 3365 passed, 1 pre-existing skip
- [x] `LocalizationParityTest` — all 16 locales carry identical `mail.post_at_risk.*` and `notifications.post_at_risk.*` keys
- [x] Per-account dedup verified via mock call-count assertions (`verify()` called once even with N posts on one account)
- [x] Cross-workspace isolation verified (a second workspace's at-risk data never leaks into the first's notification)
- [x] Concurrent-discovery race, deleted-account-mid-run, and Telegram/Discord false-negative scenarios each covered by a test confirmed to fail without its corresponding fix
- [x] `php artisan schedule:list` confirms `social:check-upcoming-connections` registered every 15 minutes alongside the two sibling jobs
- [x] `npx vue-tsc --noEmit` / ESLint clean on every touched frontend file (pre-existing, unrelated errors elsewhere confirmed present on `main` too)
- [x] Manually reproduced the original bug pattern against 3 real connected accounts (Threads, Facebook, Instagram) in a local session before this branch started — see #254


paulocastellano:
Pushed 283d4266: dropped the 16-locale translation layer from `PostAtRisk`. Turns out `App::setLocale()` is only ever set by the `SetLocale` web middleware (reads a cookie off the HTTP request) — every mailable here is built inside a queued job (`SendNotification`), which never goes through that middleware. So `config('app.locale')` always resolved to the static `APP_LOCALE` default regardless of the recipient, making the translations dead weight from the start. Replaced with plain English strings. (Same gap exists in the pre-existing `WorkspaceConnectionsDisconnected`/`AccountDisconnected` mailables — out of scope here, left as-is.)

PR #275, opened 2026-08-11
feat: migrate GTM tracking to a server-side container

paulocastellano:
Supersedes #274, which GitHub auto-closed (and now refuses to reopen) when its head branch was renamed via the API mid-review. Same branch, same commits, same content — just a fresh PR number.

## Summary
- **Dead code**: removes the unused `flash.conversion_event` push in `datalayer.ts`/`app.ts` — no backend code has ever set that flash key.
- **GTM conversion events move to the backend**: `sign_up`, `begin_checkout`, and `purchase` are no longer pushed to `window.dataLayer` from the browser (where ad blockers and cut-short page unloads were dropping them). They're now sent server-side to a GTM server-side container via a new `GtmServerService` + `SendServerEvent` job, mirroring the existing `PostHogService`/`SendEvent` pattern.
  - `sign_up` — `App\Actions\User\CreateUser` (skipped for invite registrations)
  - `begin_checkout` — `WelcomeController::storeReferralSource`
  - `purchase` — `StripeEventListener`, via a new `TrackPurchase` job reading conversion value/currency/interval from the Stripe subscription webhook payload (the conversion `transaction_id` is now the Stripe **subscription id** rather than the old Checkout Session id — flagging this in case downstream ad platforms expect the old id)
- **Config**: `GTM_ID` → `GTM_FRONTEND_ID` (⚠️ breaking rename — anyone with `GTM_ID` set needs to update it). New `GTM_BACKEND_ENABLED` / `GTM_BACKEND_ENDPOINT` / `GTM_BACKEND_API_SECRET`, independently gated so self-hosted installs can run frontend-only, backend-only, both, or neither.
- **Frontend context untouched**: the GTM container script (`gtm.blade.php`/`gtm-noscript.blade.php`) and the `dataLayer` context pushes (`user_id`, `account_id`, `account_plan`, `workspace_id`) still run client-side — tags like Crisp that depend on them keep working as-is. `useTracking.ts` keeps its PostHog `captureEvent` calls; only the `dataLayer.push` calls for the 3 named events were removed.

## Test plan
- [x] New Pest coverage: `tests/Feature/Jobs/Gtm/{SendServerEventTest,TrackPurchaseTest}.php`, plus assertions added to `CreateUserTest`, `WelcomeControllerTest`, `StripeEventListenerTest` for enabled/disabled/edge-case behavior.
- [x] Full suite: `php artisan test --compact --parallel` — 3461 passed.
- [x] `vendor/bin/pint --dirty` clean.
- [x] `vue-tsc --noEmit` / `eslint` clean on touched frontend files.
- [ ] End-to-end verification against a real GTM server-side container once one is provisioned (endpoint/payload format assumed as a simple JSON POST `{event, distinct_id, properties}` — adjust `SendServerEvent` if the container's client expects a different shape).


paulocastellano:
Descartando esse approach — decidimos usar as Destinations nativas do PostHog (Meta Ads Conversions API, Google Ads Conversions, LinkedIn Ads) em vez de rodar um container GTM server-side. Como já mandamos tudo pro PostHog via backend, resolve sem custo de infra e sem esse código.

PR #284, opened 2026-08-17
fix: API/MCP returns raw translation key when editing a finalized post

### LINKED ISSUE #244: API/MCP returns raw translation key when editing a finalized post
## Bug

When a client tries to edit/update a post that is already finalized (`published`, `publishing`, `partially_published`, or `failed`), the API and MCP return a broken error message: the **raw translation key** instead of the human-readable string.

### Observed

```json
{ "message": "posts.cannot_edit_finalized" }
```

### Expected (English)

```json
{ "message": "This post has already been processed and cannot be re-published. Duplicate it to try again." }
```

## Root cause

`App\Support\PostStatusRules::editBlockedMessage()` uses:

```php
private const EDIT_BLOCKED_MESSAGE_KEY = 'posts.cannot_edit_finalized';
```

But the string only exists under **`posts.flash.cannot_edit_finalized`** in `lang/*/posts.php`.

- Web UI already uses the correct key: `__('posts.flash.cannot_edit_finalized')` (e.g. `PostController`).
- API (`Api\PostController`) and MCP (`UpdatePostTool`, `PublishPostTool`) go through `PostStatusRules::editBlockedMessage()` and hit the wrong key.

Verified locally:

| Call | Result |
|------|--------|
| `__('posts.cannot_edit_finalized')` | `posts.cannot_edit_finalized` (missing) |
| `__('posts.flash.cannot_edit_finalized')` | Correct English copy |
| `PostStatusRules::editBlockedMessage()` | Raw key |

## Suggested fix

Either:

1. Change `EDIT_BLOCKED_MESSAGE_KEY` to `posts.flash.cannot_edit_finalized`, **or**
2. Add a top-level `posts.cannot_edit_finalized` key (and keep flash as an alias), so API/MCP and flash stay in sync.

Also update any MCP/API tests that assert `__('posts.cannot_edit_finalized')` so they assert the resolved human string (or the corrected key).

## Affected surfaces

- `PUT /api/posts/{post}` when the post is finalized
- MCP `update-post-tool` / `publish-post-tool` when the post is finalized

HafizMMoaz:
## Summary

`PostStatusRules::editBlockedMessage()` referenced the translation key
`posts.cannot_edit_finalized`, which does not exist in `lang/*/posts.php`.
The string only exists under `posts.flash.cannot_edit_finalized`, which
the web UI (`PostController`) already uses correctly. The public API and
MCP tools (`UpdatePostTool`, `PublishPostTool`, `AttachExistingAssetTool`,
`AttachExistingAsset` action) went through `editBlockedMessage()` and
returned the raw, untranslated key instead of the human-readable message.

## Changes

- Point `PostStatusRules::EDIT_BLOCKED_MESSAGE_KEY` at
  `posts.flash.cannot_edit_finalized` so it resolves to the same string
  the web UI shows.
- Update the two assertions in `PostPublishToolTest` that hardcoded
  `__('posts.cannot_edit_finalized')` (which was just echoing back the
  same missing key, so they passed without verifying anything) to assert
  the corrected key instead.

## Test plan

- [x] `php artisan test --compact tests/Feature/Mcp/PostPublishToolTest.php tests/Feature/Mcp/AssetToolTest.php tests/Feature/Api/PostMediaApiTest.php` — 84 passed
- [x] `vendor/bin/pint --dirty --format agent` — passed

Closes #244

### DIFF
--- a/app/Support/PostStatusRules.php
+++ b/app/Support/PostStatusRules.php
@@ -14,7 +14,7 @@ use Illuminate\Validation\Rule;
  */
 class PostStatusRules
 {
-    private const EDIT_BLOCKED_MESSAGE_KEY = 'posts.cannot_edit_finalized';
+    private const EDIT_BLOCKED_MESSAGE_KEY = 'posts.flash.cannot_edit_finalized';
 
     /**
      * Statuses where the post can no longer be edited.
--- a/tests/Feature/Mcp/PostPublishToolTest.php
+++ b/tests/Feature/Mcp/PostPublishToolTest.php
@@ -116,7 +116,9 @@ test('update post rejects posts in any terminal state', function (PostStatus $st
     $response = TryPostServer::actingAs($this->user)
         ->tool(UpdatePostTool::class, ['post_id' => $post->id, 'content' => 'x']);
 
-    $response->assertHasErrors([__('posts.flash.cannot_edit_finalized')]);
+    $message = __('posts.flash.cannot_edit_finalized');
+    expect($message)->not->toBe('posts.flash.cannot_edit_finalized');
+    $response->assertHasErrors([$message]);
 })->with([
     PostStatus::Published,
     PostStatus::PartiallyPublished,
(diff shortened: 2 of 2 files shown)


HafizMMoaz:
Good catch - fixed in the latest commit: now asserting the resolved string differs from the raw key before using it in assertHasErrors(), so this test actually fails if the key breaks again.

Issue #264, opened 2026-08-09
Cleanup abandoned chunked-upload sessions (orphaned temp files + open S3/R2 multipart uploads)


paulocastellano:
## Context

PR #263 gave every chunked-upload attempt a unique server-side identifier (`X-Upload-Id`, a fresh UUID per attempt) to fix a collision bug (Nightwatch issue #23). Side effect: a retry of the same file no longer reuses/continues the previous attempt's session the way it did before that fix — each attempt is now fully independent.

This means an attempt that's abandoned (user closes the media picker mid-upload and never reopens it, network drops permanently, etc.) is no longer implicitly cleaned up by a later retry. Two places accumulate orphaned state:

## 1. Local/public disk — orphaned temp files

`app/Services/Media/ChunkedAssetReceiver.php:83` writes chunks to:
```php
$tempFile = storage_path("app/private/chunks/{$identifier}");
```
It's only deleted on full completion (`ChunkedAssetReceiver.php:98`, inside a `finally` after `addMediaFromPath`). An abandoned attempt leaves its partial file in `storage/app/private/chunks/` forever — no TTL, no scheduled cleanup anywhere in the codebase.

**Suggested fix:** a scheduled Artisan command (e.g. `chunks:prune`) that deletes files in `storage/app/private/chunks/` older than some threshold (a few hours — matches `ChunkedCloudUploader::CACHE_TTL_HOURS` for consistency), registered in the scheduler.

## 2. S3/R2 — open multipart uploads never aborted

`app/Services/Media/ChunkedCloudUploader.php` — `abortIfPresent()` (`:199`) used to get triggered on a later request that reused the same identifier (a retry of the same file, pre-#263). Post-#263, since every attempt has a unique identifier, there's no longer a "later request with a matching identifier" to trigger it for an abandoned attempt. The Redis cache key expires after `CACHE_TTL_HOURS` (6h, `:22`) but that only removes our bookkeeping — it does **not** call S3's `AbortMultipartUpload`, so the incomplete upload (and its uploaded parts) stays open and billable on the bucket indefinitely.

**Suggested fix (no app code needed):** add an `AbortIncompleteMultipartUpload` lifecycle rule on the R2/S3 bucket(s) used for the `medias/` prefix, with a short expiration (1-2 days is standard). This is bucket configuration, not application code.

## Scope
- [ ] Scheduled command to prune orphaned files in `storage/app/private/chunks/`
[...]

PR #285, opened 2026-08-17
fix: prune orphaned chunked-upload temp files

HafizMMoaz:
## Summary

PR #263 gave every chunked-upload attempt a unique server-side
identifier, fixing a collision bug but also meaning a retry no longer
implicitly cleans up an abandoned attempt's session. This PR addresses
the first item in the issue's scope: orphaned local temp files.

## Changes

- Widen `ChunkedCloudUploader::CACHE_TTL_HOURS` from `private` to
  `public` so it can be reused as the single source of truth for the
  abandoned-attempt window.
- Add `chunks:prune` (`App\Console\Commands\Media\PruneAbandonedChunks`)
  which deletes files in `storage/app/private/chunks/` older than
  `CACHE_TTL_HOURS` (6h).
- Schedule it hourly in `routes/console.php`, alongside the other
  scheduled maintenance commands.

## Out of scope (infra, not app code)

The issue's second item - an `AbortIncompleteMultipartUpload` lifecycle
rule on the production R2/S3 bucket(s) for the `medias/` prefix - is
bucket configuration outside this repo. Flagging it back on the issue
so it isn't lost.

## Test plan

- [x] New `tests/Feature/Media/Command/PruneAbandonedChunksTest.php` (3 tests): prunes files older than the TTL, leaves recent files alone, no-ops when the directory doesn't exist
- [x] `php artisan test --compact tests/Feature/Media/Command/PruneAbandonedChunksTest.php tests/Feature/ChunkedAssetReceiverTest.php tests/Feature/ChunkedCloudUploadTest.php tests/Feature/ChunkedUploadFilenameEncodingTest.php` - 48 passed

Addresses #264 (app-code portion; bucket lifecycle rule still needed separately)

### DIFF
--- a/app/Console/Commands/Media/PruneAbandonedChunks.php
+++ b/app/Console/Commands/Media/PruneAbandonedChunks.php
@@ -27,8 +27,17 @@ class PruneAbandonedChunks extends Command
         $pruned = 0;
 
         foreach (glob("{$directory}/*") ?: [] as $file) {
-            if (is_file($file) && filemtime($file) < $threshold) {
-                unlink($file);
+            if (! is_file($file)) {
+                continue;
+            }
+
+            $modifiedAt = @filemtime($file);
+
+            if ($modifiedAt === false || $modifiedAt >= $threshold) {
+                continue;
+            }
+
+            if (@unlink($file)) {
                 $pruned++;
             }
         }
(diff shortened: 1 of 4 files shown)


HafizMMoaz:
Good catch - fixed: filemtime()/unlink() results are now checked explicitly rather than relying on PHP's loose false-coercion in the comparison, so an unreadable mtime or failed delete skips the file instead of proceeding incorrectly.

Issue #200, opened 2026-07-23
Harden media constraints: backend validation, reliable video meta, square YouTube Shorts


paulocastellano:
## Context

Follow-up from [#199](https://github.com/trypostit/trypost/pull/199) (YouTube Shorts max duration raised to 3 minutes in the dashboard).

Today, **duration / aspect ratio / file-size limits live only in the frontend** (`useMediaRules.ts` → `useMedia.ts`). API and MCP share `ContentTypeCompatibleWithMedia`, which only checks media **type** (image/video/document, requires media, mixed media) — not duration, dimensions, or bytes.

Media is **post-level and shared** across platforms (one `posts.media` array). Validation must stay **per `content_type` against that shared set**, not a single global limit on the file.

### 1. Reliable video meta on upload
Chunked uploads often store videos with empty `meta`. Duration/width/height used by the editor are client-supplied (or missing). Without a server probe (e.g. ffprobe), backend duration/aspect checks are forgeable or skipped.

- Extract `duration`, `width`, `height` (and maybe size) when media is finalized on the server
- Persist on asset / `post.media[].meta` so web, API, and MCP see the same values

### 2. Backend validation (web + API + MCP)
Extend `ContentTypeCompatibleWithMedia` (or a sibling rule) to enforce the same constraints as `CONTENT_TYPE_RULES` in `useMediaRules.ts`:

- max duration (`maxVideoDurationSec`)
- aspect ratio range (`aspectRatioMin` / `aspectRatioMax`)
- max file size (`maxVideoBytes` / `maxImageBytes` / …)
- file count / requires-media (where not already covered)

Apply on schedule/publish paths for:

- Web (`UpdatePostRequest`)
- Public API
- MCP (`Create`/`Update`/`Publish` post tools)

**Shared-upload model:** evaluate each enabled platform’s `content_type` independently against the shared media (same as `usePostCompliance` on the frontend). Reject the failing platform(s) / block schedule when any selected platform is incompatible — do not invent per-platform media attachments.

Prefer a **single source of truth** for the numeric rules (PHP mirror of `useMediaRules`, or a shared definition both sides read) so dashboard and API/MCP don’t drift again.

**TikTok:** max duration is account-dynamic via `creator_info.max_video_post_duration_sec` — needs a deliberate approach (fetch at validate time vs leave TikTok frontend-only for now).

**Missing meta:** decide fail-open vs fail-closed once probe exists; until then, don’t rely on client-only duration.

[...]

PR #287, opened 2026-08-17
fix: harden backend media constraints and allow square YouTube Shorts

HafizMMoaz:
## Summary

Covers goals 2 and 3 from the issue. Goal 1 (server-side video meta
via a probe like ffprobe) is deliberately left out - see below.

### Backend validation (web + API + MCP)

`ContentTypeCompatibleWithMedia` only checked media **type**
compatibility (image/video/document, mixed media, requires-media) even
though `ContentType::mediaRules()` already exposes the full numeric
rule set the Vue editor enforces via `useMediaRules`/`useMedia`
(`max_files`, `min_files`, byte caps, max video duration, aspect ratio
bounds, GIF support). Scheduling or publishing an over-limit post via
API/MCP silently skipped all of that.

Extended the rule to also check, per platform's `content_type` against
the shared post-level media (same model as `usePostCompliance` on the
frontend - one platform's failure doesn't invalidate a shared image
that's fine for a different enabled platform):

- file count bounds (`max_media_count` / `min_media_count`)
- GIF acceptance
- per-item byte size (image/video/document)
- max video duration
- aspect ratio window (skipping images on content types that
  auto-fit, e.g. Instagram Story)

Single source of truth stays `ContentType::mediaRules()` - no numbers
duplicated.

**On goal 1 (reliable server-measured video meta):** duration/width/height
come from client-supplied `media.*.meta`, same as the frontend already
uses. There's no server-side probe (ffprobe or similar) in this codebase,
and adding one means a new binary/package dependency, which needs a
separate decision. Until then, a missing meta value is treated as
unknown and skipped rather than rejected - identical to the frontend's
existing `?? 0` fallback behavior, so this doesn't change what's
enforceable today, just closes the API/MCP gap for whatever the client
already sends. Also out of scope: TikTok's account-dynamic max duration
(`creator_info.max_video_post_duration_sec`), which the issue itself
flags as needing a deliberate fetch-at-validate-time design.

### Square YouTube Shorts

YouTube accepts Shorts that are square or vertical up to 3 minutes;
`aspectRatioBounds()` only allowed ~9:16 (0.5-0.6) for `youtube_short`.
Widened to `[0.5, 1.0]` so square (1:1) is accepted, and updated the
content-type description copy.

## Test plan

- [x] `tests/Unit/Rules/ContentTypeCompatibleWithMediaTest.php` - 12 new tests: file count bounds, GIF rejection, per-type size limits, video duration, aspect ratio (including the youtube_short square case and Instagram Story's auto-fit skip), and confirms missing meta is skipped rather than rejected
- [x] `tests/Unit/Enums/ContentTypeTest.php` - new test for the widened YouTube Shorts bounds
- [x] `php artisan test --compact tests/Unit/Rules/ContentTypeCompatibleWithMediaTest.php tests/Feature/UpdatePostRequestTest.php tests/Feature/Api/PostApiPlatformMetaTest.php tests/Feature/Mcp/PostPlatformMetaToolTest.php tests/Unit/Support/PostMediaRulesTest.php tests/Unit/Enums/ContentTypeTest.php` - 80 passed, no regressions in existing type-only checks

Addresses #200 (goals 2 and 3; goal 1's server-side probe is a separate follow-up needing a dependency decision)


paulocastellano:
Thanks for this, and sorry for the slow review — it turned into a bigger investigation than the PR deserved to trigger.

Short version: the direction is right, but we found that enforcing these numbers on the backend surfaces a problem the frontend was hiding. Three of the six aspect-ratio windows don't match what the platforms actually document — Instagram Reels accepts `0.01:1` to `10:1`, not the `0.5–0.6` we had — and they're harmless today only because nothing ever measures video metadata. The duration and video aspect checks in this PR can't fire: `HasMedia` only measures images, so `meta.duration` is never populated.

So merging as-is would either do nothing for video, or — once we add a server-side probe — start rejecting Reels that publish fine today.

I wrote the whole thing up in #304, including the platform docs for each number and a task-by-task plan. **Phase 1 there is essentially this PR's scope, corrected** — split the aspect bounds by media type, fix the numbers against the docs, make the crop-aware case work, and the i18n you started in `lang/en` across all 16 locales.

Two things worth calling out that came from your PR and are now in the plan: `errorsFor()` returning the *last* violation instead of the first, and `toListingArray()` not advertising the aspect bounds it enforces — an API client gets a 422 for a rule it can't discover.

Would you like to take Phase 1? It's fully specced down to the test code, and your commits stay at the base of it. If you'd rather not, no problem at all — say the word and I'll pick it up and credit you on the PR.

PR #288, opened 2026-08-17
fix: scheduled prune for hosted social-media derivatives

### LINKED ISSUE #121: Scheduled prune for hosted social media derivatives (all networks)
## Context

Some social networks **pull media from a URL** instead of accepting an uploaded byte stream (TikTok, Instagram, Threads, Facebook). For those, when we need a processed copy of an image (aspect-ratio crop, or a resized/spec-compliant version), we render a derivative, host it on our public disk, and hand the platform that URL.

Today there are **two** derivative directories:

| Directory | Written by | Used for | Cleanup today |
|---|---|---|---|
| `social-crops/` | `App\Services\Social\Concerns\CropsImageForAspectRatio` (`:42`) | Instagram/Facebook aspect-ratio crops | **None — leaks forever** |
| `social-tiktok-photos/` | `App\Services\Social\TikTokPublisher::renderCompliantPhoto` (`:333`) | TikTok photos resized to ≤1080px (PR #120) | Inline `Storage::delete` in a `finally` after publish |

Neither approach is right long-term:

1. **`social-crops/` leaks.** Every Instagram/Facebook post with an aspect ratio writes a JPEG that is never deleted. These accumulate on disk indefinitely.
2. **`social-tiktok-photos/` inline prune has a rare timing risk.** The derivative is deleted in a `finally` right after `waitForPublishStatus` returns. On the normal `PUBLISH_COMPLETE` path (and on failure paths) TikTok has already pulled the image, so it's safe. But if the status poll times out (~60s) while TikTok is still pulling, the inline delete removes an image the platform still needs.

## Proposal

Replace the per-publisher cleanup strategy with a **single scheduled prune command** that periodically deletes derivative files older than a safe window (e.g. 1 hour — far longer than any platform needs to pull) from **all** social-derivative directories.

This is intentionally generic so that **every current and future social network** that hosts a pull-from-URL derivative is covered by one mechanism.

### Tasks

- [ ] Create `php artisan make:command PruneSocialMediaDerivatives` (sibling of the existing scheduled commands: `ProcessScheduledPosts`, `RefreshExpiringTokens`, `CheckSocialConnections`).
  - Delete files older than a configurable age (default 1h) from the known derivative directories: `social-crops/`, `social-tiktok-photos/`.
  - Source the directory list from a single place (e.g. a shared constant/config) so adding a new network's directory is one line.
- [ ] Register it in the scheduler (`->hourly()` or similar) in `routes/console.php` / `bootstrap/app.php`.
- [ ] Remove the inline `finally` prune in `TikTokPublisher::publishPhotos` (`:278`) now that cleanup is centralized — this also removes the timeout risk above.
- [ ] Tests: command deletes old files, keeps recent ones, and is a no-op when the directories are empty/missing.

### Notes / future-proofing

- When a new pull-from-URL platform starts hosting derivatives (e.g. if we ever generalize the TikTok resize to Instagram/Threads/Facebook), it must write into a directory registered with this prune, not invent an uncovered one.
- Use the application default disk (never force a disk name), consistent with the existing publishers.

### References

- PR #120 (TikTok photo resize) — introduced `social-tiktok-photos/` and the inline prune.
- `app/Services/Social/Concerns/CropsImageForAspectRatio.php` — `social-crops/`, the pre-existing leak.

HafizMMoaz:
## Summary

`social-crops/` (aspect-ratio crops and story-canvas fits, written by
`CropsImageForAspectRatio`) had no cleanup at all - every Instagram/Facebook
post with an aspect ratio leaves a JPEG on disk forever.

`social-tiktok-photos/` already has its own cleanup
(`TikTokPhotoDerivativeCleaner`), but it was added later (to fix a
resume/duplicate-publish issue) and is checkpoint-aware about in-flight
publishes - it's not a scheduled backstop, so a derivative that survives
a crash mid-publish has nothing else cleaning it up either.

## Changes

- New `social:prune-derivatives` command (scheduled hourly in
  `routes/console.php`) deletes files older than 1h - well past any
  platform's pull window - from every directory a publisher hosts a
  pull-from-URL derivative in.
- Directory names now live in one place -
  `App\Support\Social\SocialMediaDerivativeDirectory` - referenced by
  both writers (`CropsImageForAspectRatio`, `TikTokPhotoDerivativeCleaner`)
  and the new command, so a future network's directory is one line to
  add, not a hunt across publishers.

## What I intentionally left alone

The issue's task list said to remove the inline TikTok cleanup, citing
a timing risk if the status poll times out while TikTok is still
pulling the image. That's no longer accurate - since this issue was
filed, `TikTokPhotoDerivativeCleaner` was introduced with
`cleanupUnlessPublishInFlight()`, which checks a resumable publish
checkpoint before deleting anything. That's a real fix for the race,
and it runs immediately after publish (success or failure) rather than
waiting up to an hour, so removing it would be a regression, not a
cleanup. The scheduled prune here is a safety net alongside it (and
the only cleanup `social-crops/` has ever had), not a replacement.

## Test plan

- [x] New `tests/Feature/Social/Command/PruneSocialMediaDerivativesTest.php` (3 tests): prunes derivatives older than the window from both directories, leaves recent ones alone, no-ops when nothing exists yet
- [x] `php artisan test --compact tests/Feature/Social/Command/PruneSocialMediaDerivativesTest.php tests/Feature/Social tests/Unit/Support/Social` - 141 passed, no regressions in Instagram/Facebook (crop trait) or TikTok publisher tests

Closes #121

### DIFF
--- a/app/Console/Commands/Social/PruneSocialMediaDerivatives.php
+++ b/app/Console/Commands/Social/PruneSocialMediaDerivatives.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Console\Commands\Social;
+
+use App\Support\Social\SocialMediaDerivativeDirectory;
+use Illuminate\Console\Attributes\Description;
+use Illuminate\Console\Attributes\Signature;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Storage;
+
+#[Signature('social:prune-derivatives')]
+#[Description('Delete hosted social-media image derivatives (aspect-ratio crops, TikTok resized photos) once no platform still needs to pull them')]
+class PruneSocialMediaDerivatives extends Command
+{
+    private const MAX_AGE_HOURS = 1;
+
+    /**
+     * Every directory a publisher hosts a pull-from-URL derivative in. Add a
+     * new network's directory here - nowhere else - when it starts hosting
+     * derivatives a platform fetches asynchronously.
+     *
+     * @var list<string>
+     */
+    private const DIRECTORIES = [
+        SocialMediaDerivativeDirectory::CROPS,
+        SocialMediaDerivativeDirectory::TIKTOK_PHOTOS,
+    ];
+
+    public function handle(): int
+    {
+        $threshold = now()->subHours(self::MAX_AGE_HOURS)->getTimestamp();
+        $pruned = 0;
+
+        foreach (self::DIRECTORIES as $directory) {
+            foreach (Storage::files($directory) as $path) {
+                $modifiedAt = Storage::lastModified($path);
+
+                if ($modifiedAt !== false && $modifiedAt < $threshold && Storage::delete($path)) {
+                    $pruned++;
+                }
+            }
+        }
+
+        $this->info("Pruned {$pruned} social media derivative(s).");
+
+        return self::SUCCESS;
+    }
+}
--- a/app/Console/Commands/Social/PruneSocialMediaDerivatives.php
+++ b/app/Console/Commands/Social/PruneSocialMediaDerivatives.php
@@ -10,30 +10,19 @@ use Illuminate\Console\Attributes\Signature;
 use Illuminate\Console\Command;
 use Illuminate\Support\Facades\Storage;
 
-#[Signature('social:prune-derivatives')]
+#[Signature('social:prune-derivatives {--hours=1 : Delete derivatives older than this many hours}')]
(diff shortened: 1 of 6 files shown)


HafizMMoaz:
Good calls, both fixed: SocialMediaDerivativeDirectory::ALL is now the canonical list the command iterates (no more duplicate list to keep in sync), and --hours=1 makes the retention window adjustable without a code deploy.

Issue #213, opened 2026-07-29
Support Instagram `collaborators` on Reels/feed posts


kamol1dn:
### What

Instagram's Content Publishing API accepts a `collaborators` parameter when creating a media container. TryPost doesn't expose it, so posts published through TryPost can't invite collaborators — which for us means doing it by hand on every single Reel.

### Why it matters

We publish a podcast (On The Ground) and our client requires **both co-hosts to be invited as collaborators on every Instagram Reel**. It's a condition of the engagement, not a nice-to-have.

We schedule ~14 clips a week through TryPost. Every one of them currently needs someone to open the Instagram app after it publishes and re-invite the same two accounts. That's 14 manual interventions a week on posts that are otherwise fully automated — and it's the single reason our pipeline isn't hands-off.

Collaborator posts also surface on each collaborator's profile and feed, so this is the difference between one account's reach and three. Any team posting on behalf of a multi-host show or a brand partnership hits the same wall.

### The API shape

Container creation takes an array of usernames:

```http
POST https://graph.facebook.com/{ig-user-id}/media
  media_type=REELS
  video_url=...
  caption=...
  collaborators=["username1","username2"]
```

Supported on **Reels, single images and carousels**. Not supported on Stories.

Notes from Meta's docs worth encoding:

- Max 3 collaborators per post.
- Private accounts can't be tagged as collaborators via the API.
- Invited users get a request and must accept; until they do, the tag isn't public. So a successful publish does **not** mean the collaborator is live — worth reflecting in any UI copy so users aren't confused.
- There's a companion endpoint that returns invite status (`Accepted` / `Pending` / `Declined`), which would make a nice follow-up but isn't needed for v1.

### Suggested scope

Smallest useful version — accept the field and pass it through:

1. `app/Support/PostPlatformMetaRules.php` — add an Instagram key alongside the existing `aspect_ratio`. Something like:

   ```php
   // Instagram / Facebook
   'platforms.*.meta.aspect_ratio' => ['sometimes', 'nullable', 'string', Rule::enum(AspectRatio::class)],
   'platforms.*.meta.collaborators' => ['sometimes', 'nullable', 'array', 'max:3'],
   'platforms.*.meta.collaborators.*' => ['string', 'max:30'],
   ```

[...]

PR #289, opened 2026-08-17
feat: support Instagram collaborators on Reels, images, and carousels

HafizMMoaz:
## Summary

Instagram's Content Publishing API accepts a `collaborators` parameter
(up to 3 usernames) on media container creation. Without it, every Reel
published through TryPost needs a manual re-invite in the Instagram app
afterward - this covers the "smallest useful version" from the issue:
accept the field and pass it through to the container creation call.

## Changes

- `PostPlatformMetaRules::rules()`: `platforms.*.meta.collaborators`
  (array, max 3, each entry a string up to 30 chars - Instagram's
  username length cap), following the exact shape suggested in the
  issue.
- `InstagramPublisher`: a `withCollaborators()` helper adds the field
  (JSON-encoded, matching Meta's expected format and the existing
  `attached_media` JSON-encoding pattern in `FacebookPublisher`) to
  whichever container actually gets published - single image, Reel, or
  the carousel's **parent** container (not per-item children, which
  don't accept it). Reads straight from `$this->postPlatform->meta`
  (same as other instance state already on the class) rather than
  threading a new parameter through every method, which also means it
  survives a carousel publish that resumes after an interruption - the
  parent container isn't created until the resume completes.
- `publishStory()` is untouched - Instagram doesn't support collaborators
  on Stories, so it's simply never added there.

## What I intentionally left out

- **Composer UI**: the issue's suggested scope includes surfacing this
  in the web composer next to Pinterest's board picker etc. I didn't
  add it - it would be the first real usage of the `TagsInput` UI
  primitive in this app (everywhere else that needs an array-of-strings
  input, like Discord mentions, uses a server-side autocomplete instead),
  and I don't have a way to verify it renders/behaves correctly in a
  browser in this environment. Shipping an unverified new UI pattern
  felt riskier than leaving it for a follow-up where someone can
  actually click through it. The feature is fully usable today via the
  public API and MCP tools.
- **Error-code mapping**: the issue suggests mapping a specific Instagram
  error (invalid/private collaborator username) to a friendlier message
  in `InstagramPublishException`. I didn't add a guessed `error_subcode`
  entry - the existing fallback already surfaces Meta's own
  `error_user_msg` when present, and I'd rather not fabricate an
  unconfirmed code mapping (this table is built from Meta's official
  error-code docs, not guesses).
- **Status-check endpoint** (invite Accepted/Pending/Declined): explicitly
  called out in the issue as a v2 follow-up, not needed here.
- **Docs site** (docs.trypost.it): out of scope for this repo.

## Test plan

- [x] `tests/Feature/Services/Social/InstagramPublisherTest.php` - 5 new tests: collaborators on a reel container, on a single-image container, omitted when unset, never sent on a story, and present only on the carousel's parent container (not child items)
- [x] `tests/Unit/PostPlatformMetaRulesTest.php`, `tests/Feature/Api/PostApiPlatformMetaTest.php`, `tests/Feature/Mcp/PostPlatformMetaToolTest.php` - rule presence, cross-platform persistence (API), and MCP create-post coverage, plus a max-3 rejection test in each entry point
- [x] `php artisan test --compact tests/Unit/PostPlatformMetaRulesTest.php tests/Feature/Api/PostApiPlatformMetaTest.php tests/Feature/Mcp/PostPlatformMetaToolTest.php tests/Feature/Services/Social/InstagramPublisherTest.php` - 105 passed

Addresses #213 (validation + publisher; composer UI and error-message mapping left as follow-ups per above)

### DIFF
--- a/app/Services/Social/InstagramPublisher.php
+++ b/app/Services/Social/InstagramPublisher.php
@@ -117,19 +117,21 @@ class InstagramPublisher
             $params['alt_text'] = $alt;
         }
 
-        $containerId = $this->createContainer($instagramId, $params, 'container');
+        $containerId = $this->createContainer($instagramId, $this->withCollaborators($params), 'container');
 
         return $this->finishContainer($instagramId, $accessToken, $containerId);
     }
 
     private function publishReel(string $instagramId, string $accessToken, ?string $content, $media): array
     {
-        $containerId = $this->createContainer($instagramId, [
+        $params = [
             'video_url' => $media->url,
             'caption' => $content,
             'media_type' => 'REELS',
             'access_token' => $accessToken,
-        ], 'reel container');
+        ];
+
+        $containerId = $this->createContainer($instagramId, $this->withCollaborators($params), 'reel container');
 
         return $this->finishContainer($instagramId, $accessToken, $containerId);
     }
@@ -233,12 +235,12 @@ class InstagramPublisher
             $this->waitForMediaProcessing($childId, $accessToken, $workflow);
         }
 
-        $carouselId = $this->createContainer($instagramId, [
+        $carouselId = $this->createContainer($instagramId, $this->withCollaborators([
             'media_type' => 'CAROUSEL',
             'caption' => $content,
             'children' => implode(',', $childContainers),
             'access_token' => $accessToken,
-        ], 'carousel container');
+        ]), 'carousel container');
 
         return $this->finishContainer($instagramId, $accessToken, $carouselId);
     }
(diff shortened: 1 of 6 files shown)


HafizMMoaz:
Good catch - fixed both. Scoped the assertions to the actual /ig_123456789/media container-creation request (excluding media_publish), so they no longer trivially pass against the status-check GET request that also lacks collaborators.

paulocastellano:
@HafizMMoaz i'm closing this pull request in favor of https://github.com/trypostit/trypost/pull/283

It's already done for code review, test, feel free to test it.

I'm planing to merge it today/tomorrow.

Issue #225, opened 2026-08-03
[Enhancement] Use separate models for text/image/audio


rallisf1:
While there are `AI_IMAGE_PROVIDER` and `AI_AUDIO_PROVIDER` environment variables, the model falls back to `AI_TEXT_MODEL`, which is not always capable of outputting image or audio.

I suggest `AI_IMAGE_MODEL` and `AI_AUDIO_MODEL` are added and used in the respected `Ai/Agents` files.

This will enable using Ollama or OpenRouter ( pending #220 ) for all AI features in self-hosted installations.

paulocastellano:
@rallisf1 im backing to home tomorrow, so I will review all your pull requests.

Sorry to delay.

HafizMMoaz:
Dug into this expecting to add AI_IMAGE_MODEL/AI_AUDIO_MODEL, but the premise doesn't match how laravel/ai actually resolves models - wanted to explain before opening a PR that does something different than requested.

config/ai.php already has a model override per provider per capability (OPENAI_IMAGE_MODEL, GEMINI_AUDIO_MODEL, OPENROUTER_IMAGE_MODEL, etc.), and app/Services/Ai/AiImageClient.php already uses it. I checked the laravel/ai vendor source directly: every provider's image/audio default falls back to a hardcoded image/audio model (e.g. OpenAiProvider::defaultImageModel() → 'gpt-image-2'), never to that provider's text model - so there's no fallback bug in code to fix. A single global AI_IMAGE_MODEL would actually be a worse fit than what's there, since model IDs don't share a format across providers.

Also worth knowing: Ollama has no image or audio driver in laravel/ai at all (confirmed in vendor - only text/embeddings), so no config change can make Ollama do image/audio generation specifically. OpenRouter does support both already, unlike Ollama.

The real gap: .env.example never mentions any of these per-provider overrides, only the AI_*_PROVIDER selection vars - so there was no way to discover them without reading config/ai.php source. Opened #290 to surface them there. Happy to add the global vars too if you'd still want them for convenience, but they'd only work cleanly with exactly one provider configured per capability.

rallisf1:
Per provider text/image/audio models makes more sense tbh. I will be able to test in 2 days, the internet sucks where I am now. Took 3 minutes to load the gh issue page :D

rallisf1:
I can confirm the per provider environment variables work.

paulocastellano:
Thank you @rallisf1

PR #290, opened 2026-08-17
docs: surface per-provider AI model override env vars

HafizMMoaz:
## Summary

Investigated this expecting to add `AI_IMAGE_MODEL`/`AI_AUDIO_MODEL`
config as suggested, but found the underlying premise doesn't hold up
against the actual `laravel/ai` package behavior - so this PR does
something narrower and, I think, more correct.

## What I found

- `config/ai.php` already defines a model override **per provider, per
  capability** - `OPENAI_IMAGE_MODEL`, `OPENAI_AUDIO_MODEL`,
  `GEMINI_IMAGE_MODEL`, `XAI_IMAGE_MODEL`, `OPENROUTER_IMAGE_MODEL`,
  `OPENROUTER_AUDIO_MODEL`, `ELEVENLABS_AUDIO_MODEL`, etc. - and
  `app/Services/Ai/AiImageClient.php` already calls `Image::of()`
  through that config.
- There's no text-model fallback bug: I checked the `laravel/ai`
  vendor source directly (`OpenAiProvider::defaultImageModel()`,
  `OpenRouterProvider::defaultImageModel()`, etc.) and each provider's
  image/audio default falls back to a **hardcoded image/audio model**
  when unset, never to that provider's text model.
- A single global `AI_IMAGE_MODEL` (as suggested) would actually be a
  worse fit than what exists - model IDs aren't portable across
  providers (`gpt-image-2` vs. `google/gemini-3.1-flash-image-preview`
  vs. an Ollama tag), so a per-provider override is the correct shape,
  not a regression to fix.
- Ollama specifically has no image or audio driver in `laravel/ai` at
  all (confirmed in vendor - only `TextProvider`/`EmbeddingProvider`
  are implemented), so no config change here can make Ollama do image
  or audio generation. OpenRouter, unlike Ollama, already implements
  both.

The actual gap: `.env.example` only shows `AI_TEXT_PROVIDER` /
`AI_IMAGE_PROVIDER` / `AI_AUDIO_PROVIDER` (which provider to route to)
and never mentions any of the per-provider model overrides that
already exist and work. A self-hoster reading only `.env.example`
would have no way to discover them without reading `config/ai.php`
source - which lines up with the issue's report even though the
mechanism it assumed isn't what's actually happening.

## Change

Added the per-provider model override variables (commented out, as
optional) right after the existing `AI_*_PROVIDER` block in
`.env.example`, with a note on Ollama's text-only support.

## Test plan

No code paths changed - this is `.env.example` documentation only.
Verified `config/ai.php`'s existing wiring and the vendor `laravel/ai`
provider classes referenced above by reading them directly rather than
assuming.

Addresses #225 - happy to also add `AI_IMAGE_MODEL`/`AI_AUDIO_MODEL` globals if you'd still prefer those for convenience, but wanted to flag they'd need to *pick* a provider's model format, which only works cleanly if there's exactly one image/audio provider configured


paulocastellano:
@HafizMMoaz i have improved your pull requested and merged it, thank you!

PR #291, opened 2026-08-17
Fix Facebook/Instagram Page fetch missing New Pages Experience pages

StoriaJames:
## Summary
- `/me/accounts` silently omits Pages under Meta's newer "New Pages Experience" / Business Portfolio model - confirmed via Meta's own Access Token Debugger against a live account where the token's granular scopes showed the Page was explicitly granted (`pages_show_list`, `pages_read_engagement`, `pages_manage_posts` all scoped to the Page's ID), yet `/me/accounts` returned an empty list. Querying the Page node directly by ID worked fine.
- This surfaces to users as "No Facebook Pages found. You need to be an admin of at least one page." even though they are.
- Added a fallback in both `FacebookController::fetchPages()` and `InstagramFacebookController::fetchPagesWithInstagram()`: when `/me/accounts` comes back empty, fall through to Business Manager's `owned_pages`/`client_pages` edges (via `/me/businesses`), using the `business_management` scope (already requested by the Instagram-via-Facebook flow; added to the Facebook flow's scope list in this PR).

## Test plan
- [x] Reproduced against a live self-hosted instance: `/me/accounts` empty for an admin's own New Pages Experience Page, confirmed via Meta's Access Token Debugger that the grant was real and page-scoped, confirmed direct page-ID query worked.
- [x] Applied the fallback, rebuilt, reconnected the same account/Page - Page now found and connected successfully, published a real post through it.
- [ ] Would appreciate a maintainer/community check against a "New Pages Experience" Page under a Business Portfolio with multiple pages, to confirm the selection flow (`selectPage`/`select`) still behaves correctly when the fallback path returns more than one Page.

### DIFF
--- a/.github/workflows/build-storia-image.yml
+++ b/.github/workflows/build-storia-image.yml
@@ -1,0 +1,41 @@
+name: Build Storia Post image
+
+# Builds our patched production image whenever storia-main changes, and
+# pushes it to our own GHCR namespace tagged with the commit SHA (plus a
+# floating `storia-latest`). Boxes running Post clients then just
+# `docker pull` instead of anyone hand-building on a box's own daemon -
+# see the "Post: social-platform OAuth setup" section of storia-hosted-apps'
+# README for how this fits into the wider deploy flow.
+
+on:
+  push:
+    branches:
+      - storia-main
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build:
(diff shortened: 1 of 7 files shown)


StoriaJames:
Closing and reopening from a branch that contains only this fix - the one I pushed here picked up some unrelated commits from our own fork's main branch by mistake. New PR incoming.

PR #292, opened 2026-08-17
Fix Facebook/Instagram Page fetch missing New Pages Experience pages

StoriaJames:
## Summary
- `/me/accounts` silently omits Pages under Meta's newer "New Pages Experience" / Business Portfolio model - confirmed via Meta's own Access Token Debugger against a live account where the token's granular scopes showed the Page was explicitly granted (`pages_show_list`, `pages_read_engagement`, `pages_manage_posts` all scoped to the Page's ID), yet `/me/accounts` returned an empty list. Querying the Page node directly by ID worked fine.
- This surfaces to users as "No Facebook Pages found. You need to be an admin of at least one page." even though they are.
- Added a fallback in both `FacebookController::fetchPages()` and `InstagramFacebookController::fetchPagesWithInstagram()`: when `/me/accounts` comes back empty, fall through to Business Manager's `owned_pages`/`client_pages` edges (via `/me/businesses`), using the `business_management` scope (already requested by the Instagram-via-Facebook flow; added to the Facebook flow's scope list in this PR).

## Test plan
- [x] Reproduced against a live self-hosted instance: `/me/accounts` empty for an admin's own New Pages Experience Page, confirmed via Meta's Access Token Debugger that the grant was real and page-scoped, confirmed direct page-ID query worked.
- [x] Applied the fallback, rebuilt, reconnected the same account/Page - Page now found and connected successfully, published a real post through it.
- [ ] Would appreciate a maintainer/community check against a "New Pages Experience" Page under a Business Portfolio with multiple pages, to confirm the selection flow (`selectPage`/`select`) still behaves correctly when the fallback path returns more than one Page.

(Reopened from #291, which accidentally included unrelated commits from our fork's own main branch - this one contains only the fix, based directly on upstream main.)

### DIFF
--- a/app/Http/Controllers/Auth/FacebookController.php
+++ b/app/Http/Controllers/Auth/FacebookController.php
@@ -32,6 +32,10 @@ class FacebookController extends SocialController
         'pages_read_engagement',
         'pages_manage_posts',
         'read_insights',
+        // Needed for the Business Manager fallback in fetchPages() - /me/accounts
+        // omits Pages under the New Pages Experience model, and reaching those
+        // via /me/businesses + owned_pages requires this scope.
+        'business_management',
     ];
 
     public function connect(Request $request): Response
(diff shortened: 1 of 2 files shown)


paulocastellano:
Thanks for this @StoriaJames — the Access Token Debugger repro is what made the diagnosis certain, and the Business Portfolio fallback was the right call.

Your two commits are carried over as-is in #301, with a review round on top:

- the portfolio lookup moved into a shared `ManagedPages` service used by both the Facebook and Instagram-via-Facebook flows, instead of the same method living in each controller;
- `/me/businesses` made fail-soft — the point Copilot raised. Outside the `try` it escaped into the callback's `catch`, so a login without `business_management` would see "could not connect" where it used to see "no pages found";
- the portfolio edges now merge with `/me/accounts` on every connect rather than only running when it comes back empty. That covers the case you flagged in your test plan: someone holding one Page by a classic role and the rest through a portfolio was auto-connected to that single Page and never offered the others;
- Pages the login can't get an `access_token` for are dropped, since connecting one produces an account that can't publish;
- tests: six unit tests on the service plus three on the connect flows.

Opened as a separate PR only because your fork is organization-owned, and GitHub doesn't let maintainers push to a PR branch from an org-owned fork. Nothing to do with the work itself.

Closing this one in favour of #301 — the fix ships as yours.

PR #299, opened 2026-08-20
Remove the X API reads that buy nothing

paulocastellano:
Nightwatch showed the X API traffic dominated by `RefreshSocialToken` — a steady stream of `GET /2/users/me` and `POST /2/oauth2/token`, all day, for every connected account. Under X's pay-per-usage pricing that first endpoint is billed.

Two reads turned out to buy nothing, for the same underlying reason: the data they fetched was already in hand. This removes both.

### The problem

`RefreshSocialToken` verified rotating-refresh platforms instead of refreshing them:

```php
if ($this->account->platform->extendsAccessTokenOnRefresh()) {
    $verifier->refreshToken($this->account);
} else {
    $verifier->verify($this->account);   // X, LinkedIn, YouTube, TikTok, Pinterest, Bluesky
}
```

On a **still-valid** token, `verify()` only calls the platform's verify endpoint and leaves `token_expires_at` untouched. The account therefore stays inside `RefreshExpiringTokens`' 30-minute window and is re-read at every 15-minute tick until the token actually dies — then, and only then, is it refreshed.

Two consequences:

1. **We paid to answer an auth question with a data endpoint.** On X the verify endpoint is `GET /2/users/me`, billed as a **"User: Read" — $0.010 per resource**. It returns a full profile; the code reads one boolean from it (`ConnectionVerifier.php:435`). It does not qualify for Owned Read pricing ($0.001): that list is closed, `users/me` isn't on it, and even the endpoints that are only qualify when `{id}` is the developer app's owner — never true for a multi-tenant app.
2. **The token was left dead for a stretch of every cycle**, because nothing refreshed it until after expiry.

### The change

**1. Refresh instead of verify.** A provider that hands back a fresh token has already confirmed the credential — it rejects a revoked one with a 4xx, which `TokenRefreshClient` maps to `TokenExpiredException`. More than that: the refresh doesn't *inspect* the access token, it **replaces** it. After it returns, the stored token is seconds old and minted by the provider. There is nothing left for a verify call to check.

**2. Record the confirmation.** `RefreshSocialToken` stamps `last_verified_at` on a successful refresh, guarded on the platform having a refresh flow and the response carrying a usable token.

**3. Let the sweep trust it.** `VerifyWorkspaceConnections` skips accounts verified within 12 hours — the pattern `VerifyUpcomingPostConnections` has used since #255, never wired into the other callers. Only `Connected` accounts are skipped: a `TokenExpired` one still needs the call, because verifying it is how it gets promoted back.

Net effect: all three `verify()` callers stop emitting billed reads for X. What remains on the bill is inherent to the product — post creates and analytics.

### Measured

Simulated 24h of the scheduler — all three scheduled commands, 96 ticks, 8 scheduled posts — against both branches, per platform, using each provider's real token TTL.

| Platform (TTL) | `main`: refresh + verify = total | branch: refresh + verify = total | Δ calls | min/day on a dead token |
|---|---|---|---|---|
| **X** (2h) | 12 + 36 = **48** | 16 + 0 = **16** | **−67%** | 180 → **0** |
| **YouTube** (1h) | 24 + 72 = **96** | 48 + 0 = **48** | **−50%** | **360** → **0** |
| **Bluesky** (2h) | 12 + 36 = **48** | 16 + 0 = **16** | **−67%** | 180 → **0** |
| **TikTok** (24h) | 1 + 3 = **4** | 1 + 0 = **1** | **−75%** | 15 → **0** |
| **Pinterest** (30d) | 0 | 0 | — | 0 → 0 |
| **LinkedIn** (60d) | 0 | 0 | — | 0 → 0 |
| **Instagram** (60d) | 1 | **1** (identical) | — | 0 → 0 |
| **Threads** (60d) | 1 | **1** (identical) | — | 0 → 0 |

With the daily sweep and pre-publish checks included, X goes from **45 billed reads/account/day ($0.45)** to **0**.

Every platform makes strictly fewer API calls and spends zero time holding an expired token. **YouTube was in worse shape than X** and nobody knew: a 1-hour Google token entered the 30-minute window every hour, so **6 hours a day** were spent on a dead token, with every publish in that stretch eating a 401 first.

X dedupes resources within a 24h UTC window, which had been absorbing some of this — but the docs call that a *soft guarantee* that "may result in resources not being deduplicated." This takes the number of chances to leak from 45 to zero.

### Three regressions this branch introduced, found and fixed

Worth reading before approving — each was introduced by an earlier commit here, and each was found by testing against `main`, not by reasoning.

| # | Regression | `main` | before fix | Fixed in |
|---|---|---|---|---|
| 1 | Rejected refresh disconnected an account whose access token still worked | `connected` | `token_expired` | 9775578 |
| 2 | Fallback judged the stale in-memory token after a lost rotation race | `connected` | `token_expired` | 48e4195 |
| 3 | Lock-skipped refresh recorded a verification with zero HTTP calls made | n/a | stamped | 8a3cfe8 |

**#1 was not an edge case.** Any X account with no stored `refresh_token` would hit it every cycle — 16×/day, forever, each time emailing the owner and failing their posts. And X's own developer community documents refresh tokens being **invalidated spuriously**, so it would also have fired on a known, recurring X quirk.

The pattern in all three: the verify-first code had accumulated protections (`refreshThenVerify`'s reload-and-retry, the lock's early return) that the new path didn't inherit.

The test written to cover #3 originally called `refreshToken()` directly rather than going through the job, so it passed while the job path was broken. It now exercises the job and asserts no HTTP call was made.

### Two pre-existing concurrency gaps, closed here

Pulled in rather than deferred: the branch raises exposure to both from 12 to 16 refreshes/account/day, and they sit in the method it modifies.

- **The lock was the same length as the request it protects.** `Cache::lock(..., 30)` against the HTTP client's 30s default read timeout, plus 10s connect, and no refresh method set its own. Bluesky is the ceiling with two sequential calls. Named the TTL (`ConnectionVerifier::REFRESH_LOCK_SECONDS`), set it past that, and a test asserts the invariant so a future slower refresh can't quietly break it.
- **`RefreshSocialToken` wasn't unique.** `RefreshExpiringTokens` re-selects an account until `token_expires_at` moves, which only happens once the job runs — so a backlogged queue stacked one job per tick, each rotating a single-use refresh_token again for nothing. Keyed by account, matching `VerifyUpcomingPostConnections`.

### Provider behaviour, verified against official docs

| Provider | Access TTL | Refresh rotates? | Matches our code |
|---|---|---|---|
| **X** | **2h**, stated explicitly | **Yes — single-use** | ✅ `defaultTokenTtlSeconds()` = 7200 |
| **Bluesky** | 2h (hardcoded) | **Yes** — `refreshJwt` is a required output field | ✅ |
| **TikTok** | 24h / refresh 365d | "may be different — use the newly-returned token" | ✅ |
| **LinkedIn** | 60d / refresh 365d | **No** — same token, original TTL kept | ✅ |
| **Google (YouTube)** | — | **No** rotation on refresh | ✅ |

This corrected an error the branch introduced: a docblock claimed **LinkedIn** single-uses its refresh token. Its docs say the opposite. Now reads **X and Bluesky**, which is what the specs describe.

Two risk numbers gained documented bounds: Bluesky's `createSession` is rate-limited per handle at **30/5min and 300/day**, far above our 16 attempts/day; and Google doesn't rotate, so YouTube's 24 → 48 refreshes carry **no single-use risk at all**.

### Residual risk

`refreshXToken` makes the HTTP call and the `$account->update()` as two steps with `$tries = 1`. A worker death in that gap loses the new pair while the provider has already invalidated the old one — the connection is unrecoverable and the user must reconnect.

**Pre-existing, and irreducible**: none of the five provider specs offers any way to acknowledge receipt of a rotated token. What this branch controls is how often we enter the window — **12 → 16 refreshes/account/day (+33%)**, now bounded to one per account per cycle regardless of queue state.

Only X and Bluesky are exposed (single-use rotation). TikTok is unchanged at 1/day; YouTube, Pinterest and LinkedIn don't rotate.

If that matters more than the ~$0.45/account/day, narrowing the lead in `RefreshExpiringTokens` from 30 to ~5 minutes brings it to ~12.5/day, at the cost of less headroom for queue backlog. Not done without a decision.

### Behaviour vs. `main`, case by case

| Case | `main` | branch |
|---|---|---|
| Healthy account | Connected | Connected |
| Refresh rejected, access token alive | Connected | Connected |
| No `refresh_token` stored, access token alive | Connected | Connected |
| Concurrent refresh rotated the pair | Connected | Connected |
| Refresh rejected, access token dead | TokenExpired | TokenExpired |
| Platform down (5xx / network) | unchanged | unchanged |
| Meta rate-limit on Instagram extension | Connected | Connected |

One case differs by design: a legacy row on a no-refresh platform (Facebook, Mastodon, Telegram, Discord) that carries a `token_expires_at`. `main` verifies it **every 15 minutes forever** — 96 wasted calls/day, since nothing ever moves its expiry. Here it's a no-op, and detection moves to the daily sweep (24h instead of 15 min). Those platforms store `null` at connect, so this only concerns hypothetical old data.

## Part 2 — the analytics double read

`XAnalytics::getMetrics` fetched the account's timeline for post ids, then looked the **same ids** up again through `GET /2/tweets`, purely to read the `public_metrics` the first request could have returned. The timeline call asked for `start_time`, `end_time` and `max_results` — never `tweet.fields`.

Both endpoints bill per Post returned. One analytics load for an account with 250 posts in range:

| | HTTP round-trips | Posts returned by the API |
|---|---|---|
| `main` | **6** | the same 250, read **twice** |
| branch | **3** | 250, once |

**This one probably doesn't save dollars, and it's worth saying so.** X deduplicates a resource within a 24-hour UTC window, so the second read of an id already read that day isn't charged again. What it buys is half the round-trips on a user-facing page, half the rate-limit consumption for the same answer, and no longer resting 250 resources' worth of billing per load on a guarantee the docs explicitly call soft ("may result in resources not being deduplicated").

Behaviour is unchanged: same totals, same 5-page ceiling, same empty result when nothing was posted in range. The two helpers collapse into one paginated pass that sums metrics as pages arrive, and the page cap becomes a named constant — it bounds what one load can cost as much as how long it takes, which a bare `5` didn't say.

## Tests

**3767 passed, 1 skipped.** Pint clean.

19 tests on `RefreshSocialToken`, 10 on `VerifyWorkspaceConnections`, and the first 3 for `XAnalytics::getMetrics` — it had none: metrics come from the timeline instead of a second lookup, the timeline request asks for `public_metrics`, and metrics accumulate across paginated pages. New coverage includes: refresh renews without a billed read; a successful refresh stamps `last_verified_at`; a refresh whose follow-up verify fails does **not**; an empty access token in a 200 does not; a lock-skipped refresh does not; a platform with nothing to refresh does not; a rejected refresh doesn't disconnect a working account; no `refresh_token` doesn't either; a rejected refresh **does** disconnect once the access token is dead too; a lost rotation race falls back to the winner's token; a rejected refresh isn't re-sent before the access token is checked; the lock outlives the slowest possible refresh; a backlogged queue can't stack duplicate jobs.

Two existing tests were rewritten to assert the new policy rather than the old one — `refresh job routes through verify (access-token-first) not refreshToken`, and `proactive refresh does NOT rotate the X refresh token while the access token still works`. The concern behind the original policy is still guarded: rotation must leave the account `Connected`.

## Not in this PR

- **Analytics can still pull up to 500 posts** (5 pages × 100) at $0.005 each, and `since`/`until` come straight off the request without validation (`AnalyticsController.php:72`), capped at 100 days inside the service. For a heavy poster on a long range that's $2.50 per UTC day, per account. Reading fewer posts changes what the numbers mean, so it's a product decision rather than a cleanup.
- **Bluesky's `token_expires_at` is hardcoded** to `addHours(2)` in `BlueskyController.php:94` and in `refreshBlueskyToken`, rather than read from the `accessJwt`'s `exp` claim. Pre-existing; the real value is unverified.
- **`Post: Create (with URL)` costs $0.20** vs $0.015 without — 13× — and is untracked. For a scheduling product where most posts carry a link, this may well be the larger share of the bill. The Developer Console's by-endpoint breakdown settles it.

### DIFF
--- a/app/Console/Commands/RefreshExpiringTokens.php
+++ b/app/Console/Commands/RefreshExpiringTokens.php
@@ -18,10 +18,10 @@ class RefreshExpiringTokens extends Command
     protected $description = 'Proactively refresh social tokens before they expire';
 
     /**
-     * Rotating refresh_token platforms only need a short lead: verify() won't
-     * rotate a still-valid token, so we catch them right before or after expiry.
-     * Extension-model platforms (Instagram/Threads) can't be refreshed once
-     * expired, so they get a much wider lead to survive queue backlog.
+     * Rotating refresh_token platforms get a short lead: RefreshSocialToken
+     * rotates on every run, so a wider window would only rotate more often for
+     * no gain. Extension-model platforms (Instagram/Threads) can't be refreshed
+     * once expired, so they get a much wider lead to survive queue backlog.
      */
     public function handle(): void
     {
--- a/app/Console/Commands/RefreshExpiringTokens.php
+++ b/app/Console/Commands/RefreshExpiringTokens.php
@@ -18,10 +18,9 @@ class RefreshExpiringTokens extends Command
     protected $description = 'Proactively refresh social tokens before they expire';
 
     /**
-     * Rotating refresh_token platforms get a short lead: RefreshSocialToken
-     * rotates on every run, so a wider window would only rotate more often for
-     * no gain. Extension-model platforms (Instagram/Threads) can't be refreshed
-     * once expired, so they get a much wider lead to survive queue backlog.
+     * Rotating platforms get a short lead — a wider window would only rotate
+     * more often. Instagram and Threads can't be refreshed once expired, so
+     * theirs is wide enough to survive queue backlog.
      */
     public function handle(): void
     {
@@ -46,10 +45,8 @@ class RefreshExpiringTokens extends Command
                 }
             });
 
-        // Accounts in the window, not jobs queued: RefreshSocialToken is unique
-        // per account, so a dispatch is discarded while another is still in
-        // flight — and a dispatched count would overstate itself during exactly
-        // the backlog an operator reads this line to diagnose.
+        // Accounts in the window, not jobs queued: the job is unique per
+        // account, so a dispatch during a backlog is silently discarded.
         $this->info("{$count} accounts due for a token refresh.");
     }
 }
--- a/app/Console/Commands/RefreshExpiringTokens.php
+++ b/app/Console/Commands/RefreshExpiringTokens.php
@@ -46,6 +46,10 @@ class RefreshExpiringTokens extends Command
                 }
             });
 
-        $this->info("Dispatched {$count} token refresh jobs.");
+        // Accounts in the window, not jobs queued: RefreshSocialToken is unique
+        // per account, so a dispatch is discarded while another is still in
+        // flight — and a dispatched count would overstate itself during exactly
+        // the backlog an operator reads this line to diagnose.
+        $this->info("{$count} accounts due for a token refresh.");
     }
 }
--- a/app/Http/Controllers/App/AnalyticsController.php
+++ b/app/Http/Controllers/App/AnalyticsController.php
@@ -72,7 +72,10 @@ class AnalyticsController extends Controller
         $since = $request->has('since') ? Carbon::parse($request->input('since')) : null;
         $until = $request->has('until') ? Carbon::parse($request->input('until')) : null;
 
-        $metrics = match ($account->platform) {
+        // A transient platform failure — the provider being down, or a token
+        // refresh this request collided with — is not a server error. Empty
+        // numbers beat a 500 on a page the user just opened.
+        $metrics = rescue(fn () => match ($account->platform) {
             Platform::TikTok => app(TikTokAnalytics::class)->getMetrics($account),
             Platform::Instagram, Platform::InstagramFacebook => app(InstagramAnalytics::class)->getMetrics($account, $since, $until),
             Platform::Threads => app(ThreadsAnalytics::class)->getMetrics($account, $since, $until),
@@ -83,7 +86,7 @@ class AnalyticsController extends Controller
             Platform::YouTube => app(YouTubeAnalytics::class)->getMetrics($account, $since, $until),
             Platform::Telegram => app(TelegramAnalytics::class)->getMetrics($account),
             default => [],
-        };
+        }, [], report: true);
 
         return response()->json(['metrics' => $metrics]);
     }
--- a/app/Http/Controllers/App/AnalyticsController.php
+++ b/app/Http/Controllers/App/AnalyticsController.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\Http\Controllers\App;
 
 use App\Enums\SocialAccount\Platform;
+use App\Exceptions\PlatformUnavailableException;
 use App\Http\Controllers\Controller;
 use App\Models\SocialAccount;
 use App\Services\Social\FacebookAnalytics;
@@ -16,6 +17,7 @@ use App\Services\Social\ThreadsAnalytics;
 use App\Services\Social\TikTokAnalytics;
 use App\Services\Social\XAnalytics;
 use App\Services\Social\YouTubeAnalytics;
+use Illuminate\Http\Client\ConnectionException;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
 use Illuminate\Support\Carbon;
@@ -72,22 +74,41 @@ class AnalyticsController extends Controller
         $since = $request->has('since') ? Carbon::parse($request->input('since')) : null;
         $until = $request->has('until') ? Carbon::parse($request->input('until')) : null;
 
-        // A transient platform failure — the provider being down, or a token
-        // refresh this request collided with — is not a server error. Empty
-        // numbers beat a 500 on a page the user just opened.
-        $metrics = rescue(fn () => match ($account->platform) {
-            Platform::TikTok => app(TikTokAnalytics::class)->getMetrics($account),
-            Platform::Instagram, Platform::InstagramFacebook => app(InstagramAnalytics::class)->getMetrics($account, $since, $until),
-            Platform::Threads => app(ThreadsAnalytics::class)->getMetrics($account, $since, $until),
-            Platform::Facebook => app(FacebookAnalytics::class)->getMetrics($account, $since, $until),
-            Platform::X => app(XAnalytics::class)->getMetrics($account, $since, $until),
-            Platform::LinkedInPage => app(LinkedInPageAnalytics::class)->getMetrics($account, $since, $until),
-            Platform::Pinterest => app(PinterestAnalytics::class)->getMetrics($account, $since, $until),
-            Platform::YouTube => app(YouTubeAnalytics::class)->getMetrics($account, $since, $until),
-            Platform::Telegram => app(TelegramAnalytics::class)->getMetrics($account),
-            default => [],
-        }, [], report: true);
+        $metrics = $this->metricsFor($account, $since, $until);
 
         return response()->json(['metrics' => $metrics]);
+    }
+
+    /**
+     * A platform being unreachable is not a server error: empty numbers beat a
+     * 500 on a page the user just opened, and a token refresh this request
+     * collided with resolves itself within seconds.
+     *
+     * Narrow on purpose. rescue() would have caught Throwable, so a defect in
+     * any metrics service would render as "this account has no activity" with
+     * nothing but a log line to say otherwise.
+     *
+     * @return array<int, array{label: string, value: int|string}>
+     */
+    private function metricsFor(SocialAccount $account, ?Carbon $since, ?Carbon $until): array
+    {
+        try {
+            return match ($account->platform) {
+                Platform::TikTok => app(TikTokAnalytics::class)->getMetrics($account),
+                Platform::Instagram, Platform::InstagramFacebook => app(InstagramAnalytics::class)->getMetrics($account, $since, $until),
+                Platform::Threads => app(ThreadsAnalytics::class)->getMetrics($account, $since, $until),
+                Platform::Facebook => app(FacebookAnalytics::class)->getMetrics($account, $since, $until),
+                Platform::X => app(XAnalytics::class)->getMetrics($account, $since, $until),
+                Platform::LinkedInPage => app(LinkedInPageAnalytics::class)->getMetrics($account, $since, $until),
+                Platform::Pinterest => app(PinterestAnalytics::class)->getMetrics($account, $since, $until),
+                Platform::YouTube => app(YouTubeAnalytics::class)->getMetrics($account, $since, $until),
+                Platform::Telegram => app(TelegramAnalytics::class)->getMetrics($account),
+                default => [],
+            };
+        } catch (PlatformUnavailableException|ConnectionException $e) {
+            report($e);
+
+            return [];
+        }
     }
 }
--- a/app/Http/Controllers/App/AnalyticsController.php
+++ b/app/Http/Controllers/App/AnalyticsController.php
@@ -80,13 +80,9 @@ class AnalyticsController extends Controller
     }
 
     /**
-     * A platform being unreachable is not a server error: empty numbers beat a
-     * 500 on a page the user just opened, and a token refresh this request
-     * collided with resolves itself within seconds.
-     *
-     * Narrow on purpose. rescue() would have caught Throwable, so a defect in
-     * any metrics service would render as "this account has no activity" with
-     * nothing but a log line to say otherwise.
+     * An unreachable platform is not a server error — empty numbers beat a 500
+     * on a page the user just opened. Narrow on purpose: catching Throwable
+     * would render a defect as "this account has no activity".
      *
      * @return array<int, array{label: string, value: int|string}>
      */
--- a/app/Jobs/RefreshSocialToken.php
+++ b/app/Jobs/RefreshSocialToken.php
@@ -35,8 +35,9 @@ class RefreshSocialToken implements ShouldQueue
     public function handle(ConnectionVerifier $verifier): void
     {
         try {
-            $verifier->refreshToken($this->account);
-            $this->recordVerification();
+            if ($verifier->refreshToken($this->account)) {
+                $this->recordVerification();
+            }
         } catch (PlatformUnavailableException $e) {
             Log::warning('Token refresh skipped: platform unavailable', [
                 'account_id' => $this->account->id,
@@ -99,8 +100,7 @@ class RefreshSocialToken implements ShouldQueue
      * Record the refresh as a verification, so the daily sweep and the
      * pre-publish check can skip their own (often billed) verify call.
      *
-     * Only a refresh that actually happened proves anything, and only one that
-     * came back with a usable token: TokenRefreshClient classifies on HTTP
+     * Only a refresh that came back with a usable token proves anything: TokenRefreshClient classifies on HTTP
      * status alone and never inspects the body, so a 200 carrying an empty
      * token would otherwise be recorded as healthy. This is deliberately not
      * done inside ConnectionVerifier::refreshToken() — refreshThenVerify()
--- a/app/Jobs/RefreshSocialToken.php
+++ b/app/Jobs/RefreshSocialToken.php
@@ -74,10 +74,13 @@ class RefreshSocialToken implements ShouldBeUnique, ShouldQueue
 
     /**
      * A rejected refresh does not on its own mean the connection is dead.
-     * Providers that single-use their refresh_token (X, LinkedIn) reject one a
-     * concurrent refresh already consumed while the current access_token keeps
-     * working, and an account with no refresh_token at all fails here without
-     * any call being made. PublishToSocialPlatform hard-fails every post for a
+     * X and Bluesky single-use their refresh_token — X issues a new one and
+     * invalidates the previous on every refresh, and refreshSession returns a
+     * mandatory new refreshJwt — so one a concurrent refresh already consumed
+     * is rejected while the current access_token keeps working. X is also
+     * documented by its own developer community to invalidate refresh_tokens
+     * spuriously. An account with no refresh_token at all fails here without
+     * any call being made at all. PublishToSocialPlatform hard-fails every post for a
      * TokenExpired account, so disconnecting on a refresh rejection alone kills
      * posts the access_token would still have published.
      *
--- a/app/Jobs/RefreshSocialToken.php
+++ b/app/Jobs/RefreshSocialToken.php
@@ -49,6 +49,16 @@ class RefreshSocialToken implements ShouldBeUnique, ShouldQueue
     {
         try {
             if ($verifier->refreshToken($this->account)) {
+                if (blank($this->account->access_token)) {
+                    // The refresh method stored whatever came back and pushed
+                    // token_expires_at forward, so the account would otherwise
+                    // leave the refresh window looking healthy while every
+                    // publish 401s on an empty token.
+                    $this->account->markAsTokenExpired('Token refresh returned an empty access token');
+
+                    return;
+                }
+
                 $this->recordVerification();
             }
         } catch (PlatformUnavailableException $e) {
@@ -58,7 +68,7 @@ class RefreshSocialToken implements ShouldBeUnique, ShouldQueue
                 'error' => $e->getMessage(),
             ]);
         } catch (TokenExpiredException $e) {
-            if ($this->accessTokenStillWorks($verifier)) {
+            if ($this->shouldTrustAWorkingAccessToken() && $this->accessTokenStillWorks($verifier)) {
                 return;
             }
 
@@ -95,9 +105,11 @@ class RefreshSocialToken implements ShouldBeUnique, ShouldQueue
         // flight, which is why ours was rejected. This instance still holds the
         // token that was rotated away, so reload before judging it — otherwise
         // the winner's healthy account gets disconnected.
-        $this->account->refresh();
-
         try {
+            // Inside the try: the account can be deleted mid-run, and this job
+            // has tries = 1, so an escaping ModelNotFoundException fails it.
+            $this->account->refresh();
+
             return $verifier->verifyAccessToken($this->account);
         } catch (TokenExpiredException) {
             return false;
@@ -113,12 +125,28 @@ class RefreshSocialToken implements ShouldBeUnique, ShouldQueue
     }
 
     /**
+     * Whether a working access token is reason enough to stay connected after a
+     * refresh was rejected.
+     *
+     * It is for platforms that rotate a refresh_token, where a rejection often
+     * just means we lost a race and the current token is fine. It is not for
+     * Instagram and Threads: their long-lived token is extended in place and
+     * cannot be renewed once it expires, so a permanently rejected extension
+     * means the connection is already doomed. Staying connected because the
+     * token still reads would tell the owner only after it dies, when
+     * reconnecting is the only option left, and would keep retrying the
+     * rejected extension every 15 minutes across the whole 24-hour lead.
+     */
+    private function shouldTrustAWorkingAccessToken(): bool
+    {
+        return ! $this->account->platform->extendsAccessTokenOnRefresh();
+    }
+
+    /**
      * Record the refresh as a verification, so the daily sweep and the
      * pre-publish check can skip their own (often billed) verify call.
      *
-     * Only a refresh that came back with a usable token proves anything: TokenRefreshClient classifies on HTTP
-     * status alone and never inspects the body, so a 200 carrying an empty
-     * token would otherwise be recorded as healthy. This is deliberately not
+     * Deliberately not
      * done inside ConnectionVerifier::refreshToken() — refreshThenVerify()
      * calls it and can still fail on the verify that follows, and a stamp
      * written there would vouch for a credential nothing ever confirmed.
@@ -126,10 +154,6 @@ class RefreshSocialToken implements ShouldBeUnique, ShouldQueue
     private function recordVerification(): void
     {
         if (! $this->account->platform->hasTokenRefreshFlow()) {
-            return;
-        }
-
-        if (blank($this->account->access_token)) {
             return;
         }
 
--- a/app/Jobs/RefreshSocialToken.php
+++ b/app/Jobs/RefreshSocialToken.php
@@ -22,11 +22,8 @@ class RefreshSocialToken implements ShouldBeUnique, ShouldQueue
 
     public int $tries = 1;
 
-    // Covers the full schedule cadence. RefreshExpiringTokens re-selects an
-    // account until its token_expires_at moves, which only happens once this
-    // job runs — so a backlogged queue would otherwise stack a job per tick,
-    // each rotating a single-use refresh_token again for nothing and widening
-    // the window where a worker death loses the pair.
+    // token_expires_at only moves once this job runs, so a backlogged queue
+    // would stack one job per tick for the same account.
     public int $uniqueFor = 900;
 
     public function __construct(public SocialAccount $account) {}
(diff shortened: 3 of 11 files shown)


paulocastellano:
## Self-review found a regression — fixed in 9775578

Reviewing the diff for side effects turned up a real one, and it was the exact failure mode the original verify-first policy existed to prevent.

**Swapping `verify()` for `refreshToken()` dropped a safety net.** A rejected refresh went straight to `markAsTokenExpired`. But a rejected refresh doesn't mean the connection is dead:

- X and LinkedIn single-use the `refresh_token`, so one a concurrent refresh already consumed comes back `400 invalid_grant` while the current `access_token` keeps working.
- An account with no `refresh_token` at all throws before any call is made.

Verified by running the same probe against both commits:

| Scenario (access_token alive) | `main` | first commit |
|---|---|---|
| `refresh_token` burned by a race | `connected` | **`token_expired`** |
| no `refresh_token` stored | `connected` | **`token_expired`** |

`PublishToSocialPlatform.php:105` hard-fails every post for a `TokenExpired` account, so this killed posts the `access_token` would have published — up to 30 minutes before the token was actually due to expire — and emailed the owner a disconnect notice for it.

**Fix:** fall back to verifying the access token before disconnecting. That's the only path in the job that reaches the billed verify endpoint, and only after a refresh has already been rejected, so the healthy path is untouched — re-ran the 24h simulation: still **0 billed reads, 0 expired minutes**. A failure that can't be attributed to the token (platform down, network blip) leaves the account alone rather than disconnecting on noise.

Cost of the fallback is bounded: a permanently broken `refresh_token` costs ~2 reads ($0.02) over the token's remaining life, then the account legitimately goes `TokenExpired` and `RefreshExpiringTokens` stops selecting it (`status = Connected` filter).

## Other things checked, no action needed

- **`SocialAccountObserver::updated()`** early-returns unless `status` changed, so stamping `last_verified_at` on every refresh fires no notifications or PostHog syncs.
- **Platforms with nothing to refresh** can't reach the new no-op path: `RefreshExpiringTokens` requires `token_expires_at IS NOT NULL`, and Facebook / InstagramFacebook / Mastodon set it to `null` explicitly at connect, while Telegram and Discord never set it at all.
- **Lock-skipped refresh** already returned before the stamp. Confirmed, and now pinned by a test.
- **Clock skew** on `last_verified_at` only ever makes the sweep skip, never disconnect.
- **`VerifyUpcomingPostConnections`** now sees `last_verified_at` written by refreshes too, so it skips more often — intended, and covered from both ends: its 40-minute throttle is already tested there, and who writes the field is tested here.

## Tests added in this round

- a rejected refresh does not disconnect an account whose access token still works
- an account with no refresh token stays connected while its access token works
- a rejected refresh **does** disconnect once the access token is dead too
- lock skipped by a concurrent refresh does not record a verification
- a platform with nothing to refresh is never recorded as verified

One existing test (`refresh job marks account as TokenExpired when refresh_token is rejected`) now also mocks the fallback `verify()` throwing, since "refresh rejected" alone is deliberately no longer sufficient to disconnect.

Full suite: **3757 passed, 1 skipped**. Pint clean.

paulocastellano:
## Deep review round — 4 real issues found and fixed

Ran a line-by-line review of `main...HEAD`. Six findings came back; I verified each against the code rather than taking them at face value. **Four were real, all in code these commits introduced.** Fixed in the latest commit.

### 1. The stamp vouched for credentials nothing confirmed — `ConnectionVerifier.php:156`

`last_verified_at` was written inside `refreshToken()`, which `refreshThenVerify()` (`:87`) also calls. There the refresh can succeed — stamp written — and the `callVerifyEndpoint()` that follows still fail.

Usually harmless: the caller marks the account `TokenExpired`, and `recentlyProvenValid()` only skips `Connected` ones. But `markAsTokenExpired()` (`SocialAccount.php:238-242`) **silently returns** when its status lock is held by a concurrent publish. The account then stays `Connected` carrying a fresh stamp, and both skip-windows wave it through — 40 min before publishing, 12 h in the daily sweep. At-risk posts publish into a dead token with no warning email.

Moved the stamp to `RefreshSocialToken`, the caller that owns the outcome.

### 2. An empty access token was recorded as healthy — `ConnectionVerifier.php:161`

`TokenRefreshClient::send()` classifies on HTTP status alone (`:60-86`) and never inspects the body, so a 200 carrying `access_token: ""` is stored as-is and was then stamped as verified.

Worth correcting the finding as reported: a *missing* key can't get that far — `access_token` is `NOT NULL` (`create_social_accounts_table.php:24`), so `$account->update()` throws a `QueryException` before the stamp line. Only the empty-string variant is reachable. Guarded on a filled token either way.

### 3. The fallback re-sent the rejected refresh token — `RefreshSocialToken.php:46`

The fallback added in 9775578 called `verify()`, which for an **already-expired** account goes straight to `refreshThenVerify()` (`ConnectionVerifier.php:44`) and re-sends the refresh_token the provider just rejected. `RefreshExpiringTokens` selects already-expired accounts (`token_expires_at <= now()+30min`), so this is a normal path, not an edge case. On Bluesky it also re-runs the `createSession` password re-auth AT Proto rate-limits per account.

Its own docblock claimed it "only reaches the verify endpoint". Added `verifyAccessToken()`, which checks the stored token and nothing else — now the code matches the claim. Test pins it at exactly one refresh call.

### 4. `last_verified_at` was read but never written by the sweep — `VerifyWorkspaceConnections.php:52`

The field's contract was asymmetric: a successful daily verify didn't feed `VerifyUpcomingPostConnections`' 40-minute trust window, so an account the sweep had just confirmed healthy still burned a fresh call minutes later. Now stamped on the success path too.

Also fixed the `RefreshExpiringTokens` docblock, which still described the verify-first behaviour removed in c0d0d57.

### Not changed

**Rotation frequency.** The finding said healthy X accounts rotate "~12×/day instead of ~0×". That's wrong — on `main` a healthy account already rotates once per 2-hour cycle, at expiry. Measured: **12/day before, 16/day after**. A 33% increase, not 0→16. The underlying concern (a worker death between the HTTP call and `$account->update()` loses the pair while X has invalidated the old one) is real but pre-existing and unchanged in kind.

### Still 0 billed reads

Re-ran the whole-scheduler simulation — all three scheduled commands, 96 ticks, 8 scheduled posts — against both branches:

| | billed `users/me`/account/day | cost | status at end |
|---|---|---|---|
| `main` | **45** | $0.45 | `connected` |
| this branch | **0** | $0.00 | `connected` |

The 45 breaks down as 36 (`RefreshExpiringTokens`) + 1 (daily sweep) + 8 (pre-publish checks).

Full suite: **3761 passed, 1 skipped**. Pint clean.

### One open question for you — worth a look before merge

`VERIFIED_WITHIN_HOURS = 12` is shorter than the sweep's own daily cadence, so for any `hasTokenRefreshFlow()` platform the sweep now **never** runs a real verify again: refreshes land every ~90 min and keep the stamp permanently fresh.

I argued earlier this is safe because `RefreshSocialToken` detects a dead credential every ~90 min instead of daily. That holds for a dead *refresh_token*. It does not cover an access token that goes bad while the refresh keeps succeeding — with #2 fixed I can't construct a concrete way for that to happen, but the daily sweep was the backstop and it's now gone. Publish-time detection still catches it, at the cost of one failed post.

Options, cheapest first:
1. **Leave it.** Accept publish-time detection as the backstop.
2. **Raise the window past the sweep interval** (e.g. 36 h). Honest semantics — "verify at most once a day" — but for X the stamp still never goes stale, so no practical change.
3. **Force a real verify every N days** via a separate `last_full_verify_at` column. At N=7 that's ~0.06 reads/account/day — against the 45 we're removing, effectively free — and restores a true periodic check.

I'd take 3 if you want a genuine backstop, 1 if you'd rather not add a column. Say which and I'll push it.

paulocastellano:
## Closing the open question: no backstop column, no periodic paid verify

I left an open question above recommending a `last_full_verify_at` column to force a real `verify()` every N days, on the grounds that the daily sweep no longer runs one for X. Withdrawing that — the premise was wrong.

**A successful refresh doesn't inspect the access token. It replaces it.**

`refreshXToken` (`ConnectionVerifier.php:198-204`) stores the `access_token` X issued in that very response. After it returns, the stored token is seconds old, minted by X, with a full TTL ahead of it. A periodic `GET /2/users/me` would be asking X to confirm a credential X had just issued — not a failure mode worth $0.06/account/day in perpetuity.

I had been treating the refresh as a *weaker* proof than a verify. It's the stronger one for rotating platforms, because it hands back a new credential rather than vouching for an old one.

### The backstop isn't gone, it moved to the free endpoint

| Event | Detected by | Latency | Cost |
|---|---|---|---|
| User revoked the app | refresh → 4xx | ~90 min | $0 |
| `refresh_token` dead | refresh → 4xx | ~90 min | $0 |
| Empty token in a 200 | `blank()` guard, `RefreshSocialToken.php:110` | immediate | $0 |
| Scope removed | `failForMissingScopes` | at publish | $0 |

The daily sweep detected the same events at $0.010 a time, **16× slower** (once a day vs. every ~90 min). It was redundancy, not safety.

`verify()` still earns its cost for platforms with **no** refresh flow — Facebook, Mastodon, Telegram, Discord hold a long-lived token whose only health signal is asking. Those never receive a refresh stamp, so the daily sweep still checks them, and none of them bill per resource read.

The one case a periodic verify would catch and a refresh wouldn't is a stored access token corrupted on our side after the write. That's a code bug — caught by tests, not by paying X forever.

### No further changes

The branch stands as-is: **0 billed reads**, no new column, and dead-account detection ~16× faster than on `main`.

Finding 3 from the review — "the sweep effectively never runs `verify()` again for `hasTokenRefreshFlow()` platforms" — is factually right and is the intended outcome, not a regression.

paulocastellano:
## One more regression found by testing the concurrency question — fixed in 48e4195

Asked directly whether the branch was safe on *keeping tokens alive*, not just on cost, I tested it instead of reasoning about it. It wasn't.

**The fallback judged the in-memory `access_token` — the one that is stale precisely when a refresh loses a race.**

X single-uses the `refresh_token`. When two refreshes overlap, the loser's token comes back `400 invalid_grant`, and its in-memory instance still holds the pair the winner already rotated away. Verifying that token 401s, and the account is marked `TokenExpired` — while the database row holds a perfectly healthy token the winner just wrote.

| Concurrent refresh already rotated the pair | `main` | before this fix |
|---|---|---|
| account status | `connected` | **`token_expired`** |

`refreshThenVerify()` (`ConnectionVerifier.php:87-93`) already handles this — it reloads and retries with whatever was persisted. The new path never goes through `refreshThenVerify`, so it inherited none of that. Fixed by reloading before judging. Failure path only; the healthy path is untouched.

This is the second regression this branch introduced in the same area, both of the same shape: the old verify-first code had accumulated protections that the new path didn't inherit. Worth weighing when reviewing.

## Where the branch stands

Measured over a simulated 24h with all three scheduled commands and 8 scheduled posts:

| | billed reads/account/day | cost | minutes on a dead token | status at end |
|---|---|---|---|---|
| `main` | 45 | $0.45 | 180 | `connected` |
| this branch | **0** | **$0.00** | **0** | `connected` |

Full suite: **3762 passed, 1 skipped**. Pint clean.

### Behaviour vs. main, verified case by case

| Case | `main` | branch |
|---|---|---|
| Healthy account | Connected | Connected |
| Refresh rejected, access token alive | Connected | Connected |
| No `refresh_token` stored, access token alive | Connected | Connected |
| Concurrent refresh rotated the pair | Connected | Connected |
| Refresh rejected, access token dead | TokenExpired | TokenExpired |
| Platform down (5xx / network) | unchanged | unchanged |
| Meta rate-limit on Instagram extension | Connected | Connected |

## Residual risk, stated plainly

`refreshXToken` makes the HTTP call and the `$account->update()` as two steps, with `$tries = 1`. A worker death in that gap loses the new pair while X has already invalidated the old one — the connection is unrecoverable and the user must reconnect.

This is pre-existing, not introduced here, but the branch raises the exposure: **12 rotations/day/account on `main` → 16 here** (+33%), because the window now opens ~30 min before expiry each cycle instead of at expiry. The gap itself is microseconds; realistically it only matters during a deploy that kills workers mid-flight.

If that matters more than the ~$0.45/account/day, narrowing the lead in `RefreshExpiringTokens` from 30 min to ~5 min brings it to ~12.5/day at the cost of less headroom for queue backlog. Not doing that without a decision.

paulocastellano:
## Both concurrency gaps closed in 5001c9a

Pulled these into the PR rather than deferring them: they're pre-existing, but this branch raises the exposure from **12 to 16 refreshes per account per day**, and they sit in the exact method it modifies.

### The lock was the same length as the request it protects

`Cache::lock("token_refresh:{$account->id}", 30)` — and 30s is precisely the HTTP client's default read timeout (`PendingRequest.php:271`), with a further 10s connect timeout on top. No refresh method sets its own.

Bluesky is the ceiling: it refreshes with two sequential calls (`refreshSession`, then the `createSession` re-auth), so up to ~80s of work under a 30s lock. Once the lock lapses mid-flight, a second process refreshes with the **same single-use refresh_token** and one of the two comes back rejected.

Named the TTL (`ConnectionVerifier::REFRESH_LOCK_SECONDS`), set it past the ceiling, and wrote the invariant into the docblock. A test asserts the relationship holds, so a future slower refresh can't quietly break it. The lock is released in a `finally`, so a longer TTL only governs how long a worker that died mid-refresh blocks the next attempt — and the next tick is 15 minutes out regardless.

### `RefreshSocialToken` wasn't unique

`RefreshExpiringTokens` re-selects an account until `token_expires_at` moves, and that only happens once the job runs. A queue more than one tick behind therefore stacked one job per tick for the same account — each rotating a single-use refresh_token again for nothing, and widening the window where a worker death loses the pair.

Keyed by account with `uniqueFor = 900`, matching `VerifyUpcomingPostConnections`. Test drives the command twice without a worker and asserts one job.

### Nothing else moved

| | billed reads/account/day | refreshes/day | minutes on a dead token | status at end |
|---|---|---|---|---|
| `main` | 45 | 12 | 180 | `connected` |
| this branch | **0** | 16 | **0** | `connected` |

Full suite: **3764 passed, 1 skipped**. Pint clean.

### On the residual risk

The non-atomic gap in `refreshXToken` — HTTP call, then `$account->update()`, with `$tries = 1` — is still there. It can't be closed without a protocol X doesn't offer: there's no way to confirm receipt of a rotated token. What these two fixes do is remove the *avoidable* ways of entering that gap more often than necessary. Frequency is now back to one refresh per account per cycle, whatever the queue is doing.

paulocastellano:
## Per-platform impact — `RefreshSocialToken` is shared by every network

`RefreshSocialToken` and `ConnectionVerifier` are shared code, and everything above measured X only. Simulated 24h of `social:refresh-expiring-tokens` for every platform against both branches, using each provider's real token TTL.

| Platform (token TTL) | `main`: refresh + verify = total | branch: refresh + verify = total | Δ calls | min/day on a dead token |
|---|---|---|---|---|
| **X** (2h) | 12 + 36 = **48** | 16 + 0 = **16** | **−67%** | 180 → **0** |
| **YouTube** (1h) | 24 + 72 = **96** | 48 + 0 = **48** | **−50%** | **360** → **0** |
| **Bluesky** (2h) | 12 + 36 = **48** | 16 + 0 = **16** | **−67%** | 180 → **0** |
| **TikTok** (24h) | 1 + 3 = **4** | 1 + 0 = **1** | **−75%** | 15 → **0** |
| **Pinterest** (30d) | 0 | 0 | — | 0 → 0 |
| **LinkedIn** (60d) | 0 | 0 | — | 0 → 0 |
| **Instagram** (60d) | 1 | **1** (identical) | — | 0 → 0 |
| **Threads** (60d) | 1 | **1** (identical) | — | 0 → 0 |

**Every platform makes strictly fewer API calls and spends zero time on a dead token.**

### YouTube was in worse shape than X

A Google access token lives 1 hour, so a YouTube account entered the 30-minute window *every hour*: 96 calls/day and **6 hours a day holding an expired token**. Every YouTube publish in that stretch ate a 401 before it worked. Now zero.

Its refresh count does go 24 → 48/day, which looks like a doubling in isolation — but the 72 verify calls disappear with it, so total traffic halves. Google doesn't rotate refresh tokens by default, so those 48 carry no single-use risk.

### Where the added rotation risk actually lands

Only on platforms whose refresh_token is single-use: **X (12→16/day)** and **Bluesky (12→16/day)**. TikTok is unchanged at 1/day, and Pinterest/LinkedIn never enter the window at all. That is the same +33% exposure to the non-atomic gap documented above, now bounded per cycle by `ShouldBeUnique`.

Bluesky deserves the extra note: its refresh falls back to a `createSession` password re-auth, which AT Proto rate-limits per account. 16 attempts/day is far below that, and the `verifyAccessToken()` change stops the fallback from firing a *second* re-auth.

### Instagram / Threads: verified unchanged, not assumed

They were already the `refreshToken()` branch on `main` — the `if` this PR removed. Simulation returns identical numbers on both branches.

### The one case where behaviour genuinely differs

Facebook, InstagramFacebook, Mastodon, Telegram and Discord can't reach the changed code: `RefreshExpiringTokens` requires `token_expires_at IS NOT NULL` and their connect flows store `null`. That's a claim about production data, so it's worth showing what happens if a legacy row does carry an expiry:

| Mastodon account with `token_expires_at` set | `main` | branch |
|---|---|---|
| API calls / 24h | **96** | **0** |

On `main` such an account is verified every 15 minutes forever — nothing ever moves its `token_expires_at`, so it never leaves the window. 96 wasted calls a day. On this branch it's a no-op: `refreshToken()` hits `default => null`, and `recordVerification()` declines to stamp because `hasTokenRefreshFlow()` is false.

Detection is not lost — with no stamp, `recentlyProvenValid()` is always false, so the daily sweep still verifies it. The trade is detection latency for that hypothetical row (15 min → 24 h) against removing 96 pointless calls a day. Worth flagging rather than burying.

paulocastellano:
## Cross-checked every provider's official OAuth docs against the implementation

The risk model above rested on assumptions about how each provider handles refresh tokens. Verified them against the official documentation instead.

| Provider | Access token TTL | Refresh token rotates? | Matches our code |
|---|---|---|---|
| **X** | **2h**, stated explicitly | **Yes — single-use, previous invalidated** | ✅ `defaultTokenTtlSeconds()` = 7200 |
| **Bluesky** | hardcoded 2h | **Yes** — `refreshSession` declares `refreshJwt` a *required* output | ✅ |
| **TikTok** | 24h / refresh 365d | "may be different — you must use the newly-returned token" | ✅ `data_get($data, 'refresh_token', $account->refresh_token)` |
| **LinkedIn** | 60d / refresh 365d | **No** — same token returned, original TTL kept | ✅ same fallback works |
| **Google (YouTube)** | — | **No** rotation on refresh | ✅ |

### Corrected a factual error this branch introduced

A docblock added in 9775578 said "providers that single-use their refresh_token (**X, LinkedIn**)". LinkedIn's docs say the opposite: *"the lifespan or Time To Live (TTL) of the refresh token remains the same as specified in the initial OAuth flow (365 days)"* — the same token comes back with a decreasing `refresh_token_expires_in`, and only the access token is reissued.

The claim predates this branch (the comment in `verify()` carried it too) but the new docblock repeated it. Both now say **X and Bluesky**, which is what the specs actually describe. Fixed in the latest commit.

### The fallback matters more than I thought

X's own developer community documents refresh tokens being **invalidated spuriously** — a recurring, long-running thread of `"Value passed for the token was invalid"` on tokens that were unquestionably valid.

That reframes the fallback added in 9775578. It isn't only race protection: without it, every one of those spurious invalidations disconnects a healthy account, emails the owner, and fails their scheduled posts. The first commit on this branch would have turned a known X quirk into a customer-visible outage.

### Two risk numbers now have documented bounds

- **Bluesky** — `createSession` is rate-limited per handle at **30 per 5 minutes and 300 per day**. The fallback re-auth path shares that budget; at 16 refresh attempts/day it sits far below, with the `verifyAccessToken()` change stopping a second re-auth per attempt.
- **YouTube** — Google doesn't rotate on refresh, so the 24 → 48 refreshes/day carries **no single-use risk at all**. Its documented limit is 100 refresh *tokens* per account per client, which concerns new authorizations, not refresh calls.

### What reading the docs could not close

The non-atomic gap in `refreshXToken` — HTTP call, then `$account->update()`. No provider in the table offers any way to acknowledge receipt of a rotated token; I looked for one in each spec. It's an irreducible property of rotating refresh tokens, not an oversight in this code. What the branch controls is how often we enter that window, now bounded to one refresh per account per cycle regardless of queue state.

### Unrelated observation worth a follow-up

`token_expires_at` for Bluesky is hardcoded to `now()->addHours(2)` in both `BlueskyController.php:94` and `refreshBlueskyToken`, rather than read from the `accessJwt`'s own `exp` claim. Pre-existing and untouched here. AT Proto material describes the access token as shorter-lived than that; I have not verified the real value, so flagging rather than asserting.

**Sources:** [X OAuth 2.0 Authorization Code Flow with PKCE](https://docs.x.com/fundamentals/authentication/oauth-2-0/authorization-code) · [LinkedIn refresh tokens](https://learn.microsoft.com/en-us/linkedin/shared/authentication/programmatic-refresh-tokens) · [TikTok token management](https://developers.tiktok.com/doc/oauth-user-access-token-management) · [AT Proto refreshSession lexicon](https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/server/refreshSession.json) · [Google OAuth 2.0](https://developers.google.com/identity/protocols/oauth2)

paulocastellano:
## Verified against the live X API, not fakes

With a real connected account in the local database, the two claims this PR rests on stopped being inference.

### Keeping a token alive costs $0.00

Ran the actual `RefreshSocialToken` job against api.x.com, logging every request the job made:

```
HTTP CALLS MADE  => ["POST /2/oauth2/token"]          ← one call, unbilled
billed users/me  => 0

access_token   VjgxTDA2WHd4 → OGVlTGdMZ0k1            ← rotated
refresh_token  Qm4yajJHM2ty → N3B5b3N6Mzcx            ← rotated (single-use confirmed)
token_expires  21:22:36     → 21:29:37                ← now + 2h
last_verified  (null)       → 19:29:37                ← stamped
status         connected                               ← intact
```

One free call, zero billed reads, token replaced rather than inspected. The refresh_token changing also confirms X's single-use rotation empirically, which until now came only from the developer forum.

### The fallback works in both directions

The safety net only runs when a refresh is rejected, so it had never been exercised against the real API — and a broken one disconnects healthy accounts:

| Input | Result |
|---|---|
| Valid access token | `true` → stays Connected, no false disconnect |
| Deliberately invalid token | `TokenExpiredException` → correctly marked TokenExpired |

It issues exactly one `GET /2/users/me`, confirming the billed call is confined to the failure path.

### Analytics end to end

```
HTTP CALLS => ["/2/users/62378351/tweets?start_time=…&end_time=…"]
bulk /2/tweets lookups => 0
metrics => Impressions 223 · Likes 3 · Retweets 1 · Replies 1 · Quotes 0 · Bookmarks 0
```

One request, real numbers, no second lookup.

### The `tweet.fields` question, settled

The two official sources contradict each other — the OpenAPI spec names the parameter `post.fields`, while the Fields guide and every example use `tweet.fields`. Tested all four cases live:

| Sent | Status | `public_metrics` returned |
|---|---|---|
| `tweet.fields=public_metrics` | 200 | **yes** |
| `post.fields=public_metrics` | 200 | **yes** |
| `bogus.fields=public_metrics` | 200 | **no** |
| nothing | 200 | **no** |

Both names are accepted. More importantly: **an unrecognised parameter returns 200 with no error and silently omits the field.** A wrong name here means an analytics page of zeros with nothing in the logs and no test able to catch it, since fakes return whatever they're told to. The name being right is load-bearing, and it is.

## One trap this branch created, now guarded

Dropping `default => null` from `refreshToken()` in af379af made the return value honest, but turned a missing case into an `UnhandledMatchError` at runtime. The arms and `hasTokenRefreshFlow()` agree on the same nine platforms today, and nothing enforced that — adding a platform to the predicate without an arm would fail in production, on a queue worker, for one platform's accounts only.

There's now a test that walks every platform claiming a refresh flow and fails by name if no arm exists. Confirmed it catches the real thing by temporarily adding Facebook to the predicate (it failed with `facebook claims a refresh flow but refreshToken() has no arm for it`) rather than trusting a green run on code that already agrees with itself.

Full suite: **3775 passed, 1 skipped.** Pint clean.

PR #309, opened 2026-08-30
YouTube descriptions and a first comment after publish (YouTube + Instagram)

gynsus:
## Why

A YouTube Short published through TryPost gets its description built from the post text — which is also what the 100-character title cap was applied to, so any post longer than a tweet failed outright, and there was no way to give a Short the long, link-bearing description (5,000 chars, clickable URLs) that YouTube SEO actually runs on. The classic "link in the first comment" trick wasn't possible either.

## What changed

- **`meta.description` for YouTube** (≤ 5,000 chars, falls back to the post content): a dedicated panel in the editor's YouTube channel settings. The video title still comes from the post content's first line; the description is now its own field end to end.
- **The content cap follows the description limit, not the title's.** `maxContentLength` for YouTube was 100 — the *title* limit — so a 372-character post died with "content exceeds limit" even though it was only ever used as a description. It is now 5,000, and the title builder trims to 100 on its own.
- **`meta.first_comment`** for **YouTube** (`commentThreads.insert`, ≤ 10,000 but capped at 2,200 to share the field with Instagram) and **Instagram** (`/{media_id}/comments`; `instagram_manage_comments` joins the requested scopes — existing IG connections need one reconnect to pick it up). Posting the comment happens *after* the post is marked published, in a `FirstCommentPoster` that catches everything and only logs: a failed comment must never fail a published post.
- Both fields validate through `PostPlatformMetaRules`, so the web editor, the REST API, and MCP all accept and persist them (fields without a rule there get silently stripped by `validated()` — covered by the meta persistence tests).

## Testing

`YouTubePublisherTest`, `FirstCommentPosterTest`, `PostApiPlatformMetaTest`, `PostPlatformMetaToolTest`, and the platform cap test pass. Battle-tested live: Shorts published with long descriptions + first comments on YouTube, first comments on Instagram.


paulocastellano:
@gynsus it's ready for review?

gynsus:
@paulocastellano Yes, ready for review! The branch is rebased on v1.1.0 (also adapted the new repurpose caption tests to the raised YouTube description limit) and CI is fully green.

PR #311, opened 2026-08-30
Add VK (VKontakte) as a platform

gynsus:
## Why

VK (VKontakte) is the dominant social network for Russian-speaking audiences, and self-hosted TryPost users publishing there currently have to do it by hand. This PR adds VK as a full platform: connect, publish (text, photos, video), and analytics.

The connect flow is deliberately **token-based, not OAuth**. VK stopped granting the `wall` scope to newly registered OAuth apps years ago, so a stock "log in with VK" integration cannot post to walls at all. What still works, verified against the live API:

- a **community access token** (community → Manage → API usage → Access tokens) — `wall.post` with `owner_id=-<group>` + `from_group=1` is allowed for these regardless of the app type that issued them; this is the recommended path and the one the form suggests first;
- a **user access token** issued by a standalone app the user owns (`wall, photos, groups, video, offline` scopes) — needed for profile walls and for video upload, which VK only allows with user tokens.

## What changed

- `Platform::Vk` + `ContentType::VkPost` (16,000-char limit, VK's actual cap), lang strings for all 16 locales, platform tile/logo/theme.
- **Connect** (`VkController`): one token field for both token kinds. A community token is detected by a successful-but-empty `users.get` (the method accepts community tokens and just returns an empty list without `user_ids` — it does *not* error). Three live-API quirks shaped the flow:
  - VK has **no API to tell which community a token belongss to**, so for community tokens the form asks for the community address (`vk.com/club…`, screen name, or bare id — all normalized) and resolves it via `groups.getById`.
  - Ownership is then proven with `groups.getCallbackConfirmationCode`, which only answers for the token's own community. (`groups.getOnlineStatus` looked right for this but fails with "group messages is disabled" for communities that turned messages off.)
  - For user tokens, `users.get` + `groups.get(filter=admin,editor)` builds a picker of publishable walls (own profile + administered communities).
- **Publish** (`VkPublisher`): `wall.post` with attachments; photos via the three-step `getWallUploadServer → upload → saveWallPhoto` flow (images optimized first), video via `video.save` with `wallpost=0` so the video attaches to our post instead of auto-publishing. A failed single attachment is skipped with a log, not a sunk post. VK reports errors as HTTP 200 + `error` object; `VkPublishException` maps codes to the shared error categories (rate limit / permission / media format).
- **Analytics** (`VkAnalytics`): community member count on the account level; views/likes/reposts/comments per post via `wall.getById`. Handles both the v5.199 response shape (`response.groups[]` / `response.items[]`) and the older flat one.
- **Verification** (`ConnectionVerifier`): `users.get` for user tokens, `groups.getById` for community tokens (which `users.get` can't verify); VK error 5 is the confirmed-dead-token signal.

## Testing

- New suites: `VkControllerTest` (9 tests: both token kinds, community mismatch/unknown address, VK-side errors), `VkPublisherTest`, `VkAnalyticsTest`, `VkPublishExceptionTest`, plus `ConnectionVerifierTest` additions for both verify paths.
- Battle-tested on a self-hosted production instance: both token kinds connected, wall posts with photos published to two communities, analytics rendering member counts and per-post metrics.

## What only you can decide

Token-based connect is unusual for this codebase — every other platform is OAuth. We kept the UI to the same two-step popup pattern as LinkedIn's identity picker, but if you'd rather gate VK behind a `SELF_HOSTED`-style env flag until it settles, that's a small change on top.


paulocastellano:
@gynsus have you checked my last email?

gynsus:
@paulocastellano Sorry for the slow reply — just saw your email. To answer the question: **no login needed at all**, not in the mobile app and not on vk.com. The community access token I sent works headlessly: paste it into the Connect VK form, enter `vk.com/club241355339` on the second step, and publish — the token itself carries the posting rights for that community. A VK account/login only becomes relevant for the optional user-token path (profile walls + video upload). Rebased the branch onto current main a few days ago, so it's conflict-free and green.

PR #315, opened 2026-08-30
Configurable TikTok OAuth scopes for apps without the Display API

gynsus:
## The problem

TikTok no longer offers the **Display API** product to new apps. That product owns three of the six scopes the controller hardcodes — `user.info.profile`, `user.info.stats`, `video.list` — so a freshly registered TikTok app (Login Kit + Content Posting API is all you can add today) fails the consent screen with a bare `scope` error, and there is no way to connect TikTok at all on a new self-hosted install.

## The fix

Requested scopes come from config:

```
TIKTOK_SCOPES=user.info.basic,video.publish,video.upload
```

with the default unchanged (all six scopes), so existing installs are unaffected. The rest of the integration already degrades gracefully when the trimmed scopes are missing: analytics show nothing for a token without `user.info.stats`, and the username simply stays empty without `user.info.profile`.

## Testing

`TikTokControllerTest` passes. Verified live on a self-hosted install with a 2026-registered TikTok app: with the trimmed scope list the consent screen renders, the account connects, and Direct Post publishing works (to a private account until the app passes TikTok's audit, per their unaudited-client policy).


paulocastellano:
@gynsus make sense it be equal linkedin, i will add it.

paulocastellano:
@gynsus Nice one — thanks for sending this in. I just tightened a few things on top (env example, tests, and a stale comment), but the change itself is solid. Appreciate the suggestion.

PR #320, opened 2026-08-31
fix(facebook): use live post insight metrics and read Reels from `/video_insights`

maciejdzierzek:
Facebook posts currently return **no metrics**, for two independent reasons in Graph API v25. Companion to #319 (same problem class on Instagram).

### 1. Feed posts ask for retired metrics

`post_impressions` and `post_impressions_unique` were retired. A retired name does not get skipped — it fails the *whole* request, so the other metrics in the same call (`post_reactions_like_total`, `post_clicks`) never arrive either.

Current equivalents:

| retired | current |
| --- | --- |
| `post_impressions` | `post_media_view` |
| `post_impressions_unique` | `post_total_media_view_unique` |

### 2. Facebook Reels have no `/insights` edge

Every Reel errored out, because Reels expose their metrics on `/video_insights` instead. Called without a `metric` parameter that edge returns a sensible default set (`blue_reels_play_count`, `post_impressions_unique`, `post_video_avg_time_watched`, and others), which is what this PR reads.

Reaction counts there come back as a dictionary keyed by reaction type, so those values are summed into a single number.

### Labels

The raw v25 metric names do not survive `ucfirst(str_replace(...))` — `post_total_media_view_unique` renders as *"Post total media view unique"*. This maps them to the same vocabulary the other analytics services already use: Reach / Views / Likes / Clicks, plus Plays / Replays / Avg watch time for Reels.

### How I verified

Probed both edges against live Business assets on v25. Calling the insights edge without a `metric` parameter returns the default set, and passing a retired name returns an error listing the permitted metrics — that is how the replacement names above were established rather than guessed.

Running in production since 2026-08-04. Before: every Facebook post reported an API error. After: reach and views populate for both feed posts and Reels.

### DIFF
--- a/app/Services/Social/FacebookAnalytics.php
+++ b/app/Services/Social/FacebookAnalytics.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace App\Services\Social;
 
+use App\Enums\PostPlatform\ContentType;
 use App\Models\PostPlatform;
 use App\Models\SocialAccount;
 use App\Services\Social\Concerns\HasSocialHttpClient;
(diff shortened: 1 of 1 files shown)


paulocastellano:
Closing — we already shipped this on `main` in #361.

That PR covers the same Facebook analytics breakages (retired `post_impressions*` on feed posts, Reels on `/video_insights`, plus Story metrics and tests). This branch now conflicts with that work and is superseded.

PR #360, opened 2026-09-17
fix(facebook): use rupload upload_url for story video transfer

Cryptoom:
## Summary

`publishStory()` ignored the `upload_url` returned by the Video Stories API's start phase and instead POSTed the media's hosted URL as a body field called `video_file_chunk` to `/{video_id}` on the regular Graph host. That is not a valid transfer step for this API: Facebook's Video Stories API always returns `{video_id, upload_url}`, and the transfer phase has to reach the rupload host at `upload_url`, not `graph.facebook.com`.

Every Facebook Story publish failed with Graph API error 6000 ("Problem with file. Try with another file."), regardless of the actual video (reproduced with and without an audio track).

## Fix

Mirrors the already-working `publishReel()` transfer step: download the hosted media and stream raw bytes to `upload_url` with the `Offset`/`file_size` headers rupload requires. The documented `file_url`-header shortcut is the same shortcut `publishReel()`'s own comment already flags as unreliable.

Verified against Meta's documented start/transfer/finish flow (cross-checked against a third-party Facebook Stories API writeup that matches the official docs).

## Test plan

- [x] `php -l` on both changed files
- [x] Updated `FacebookPublisherTest.php` mirrors the already-passing reel test's assertions against the new transfer contract
- [ ] Could not run the full local test suite on this machine (no working Postgres test database, port 5432 held by an unrelated project's container). CI should cover this.


paulocastellano:
Thank you @Cryptoom i will review it soon.

paulocastellano:
Thanks for this, @Cryptoom — your diagnosis was right: the transfer was posting `video_file_chunk` to the Graph `/{video_id}` node, so the `upload_url` session stayed empty and `finish` failed with 6000.

We went with the hosted-file variant of the same official flow in #361: `file_url` header on the `upload_url` plus a status poll before `finish`, so the worker doesn't download and re-upload the video. Reels stay as they are in `main`.

Keeping this open until #361 is verified and merged, then I'll close it.

paulocastellano:
Closing in favor of #361, which applies the fix from main with the hosted-file (file_url) flow. Thanks again for tracking down the root cause.

